PR #1, opened 2026-06-12
feat: add pi extension

HannesOberreiter:
Adds Pi agent harness support for Ponytail.

## Changes

- Add Pi package manifest via `package.json`
- Add Pi extension with `/ponytail`, `/ponytail-review`, `/ponytail-help`
- Persist active Ponytail mode per Pi session
- Inject Ponytail instructions into Pi system prompt
- Share config/instruction logic between Claude hooks and Pi extension
- Document Pi install command
- Add extension tests

Screenshot 2026-06-12 at 09 06 29

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,12 @@ codex
 Open `/plugins`, select the Ponytail marketplace, and install Ponytail. Then
 open `/hooks`, review and trust its two lifecycle hooks, and start a new thread.
 
+### Pi agent harness
+
+```
+pi install git:github.com/DietrichGebert/ponytail
+```
+
 That was it. He'd be proud. He won't say it.
 
 Active every session. `/ponytail-review` finds what to delete in your diff. `/ponytail ultra` exists for when the codebase has wronged you personally. `/ponytail-help` explains the rest.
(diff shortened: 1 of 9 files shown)


DietrichGebert:
﻿Thanks for this, the Pi adapter is thorough and the test coverage is exactly the bar I want for new adapters.

Two things before it can land:

1. It needs a rebase onto current `main`. This branch was cut before the v4 hardening and main has moved since, so it currently conflicts (the README install section and the hook fallback text both changed). After rebasing, please confirm the fallback instruction string in `ponytail-instructions.js` matches the current `skills/ponytail/SKILL.md` wording, which now carries the ceiling/upgrade-path and "one runnable check" clauses.

2. PR #7 (Codex support) just merged. It also factored the hooks into a shared module, `hooks/ponytail-runtime.js`, that switches state path and output format by host (gated on `PLUGIN_DATA`). Your `ponytail-instructions.js` split is complementary in spirit, but both touch `ponytail-activate.js`. Please reconcile against the merged `ponytail-runtime.js` so the two refactors compose instead of collide.

Once it is rebased and green, I am happy to merge. Thanks again for the work.

PR #2, opened 2026-06-12
Document Ponytail agent adapters

ChindanaiNaKub:
## Summary
- add a focused adapter table for Claude Code, Codex, Cursor, Windsurf, Cline, Copilot, Kiro, and generic agents
- document which files are runtime/plugin adapters versus compact instruction adapters
- keep the PR docs-only after #7 added Codex support

## Checks
- git diff --check
- node scripts/check-rule-copies.js
- node --test tests/hooks.test.js

No prompt behavior changes.

### DIFF
--- a/docs/agent-portability.md
+++ b/docs/agent-portability.md
@@ -1,0 +1,31 @@
+# Agent Portability
+
+Ponytail is an agent-portable skill distribution. The skills in `skills/` hold
+the core behavior; host-specific files are adapters that make that behavior easy
+to load in a given agent.
+
+## Supported Adapters
+
+| Host | Files | Notes |
+|------|-------|-------|
+| Claude Code | `.claude-plugin/`, `commands/`, `hooks/` | Full plugin install with session activation, mode tracking, commands, and statusline support. |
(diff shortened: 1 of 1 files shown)


DietrichGebert:
﻿Thanks for this. For Codex support I went with #7, which ships the same `.codex-plugin/plugin.json` plus the lifecycle-hook runtime and a cross-platform compatibility test, so the manifest here is now superseded and would conflict with main.

The piece that is still uniquely useful is the adapter table in `docs/agent-portability.md`. If you want to rework this PR down to just that doc (dropping the duplicate manifest, and the `CONTEXT.md` glossary, which is a bit more than this minimalist repo needs), I would happily take it as a focused docs PR. Otherwise I am leaning toward closing this in favor of #7. Appreciate the contribution either way.

ChindanaiNaKub:
Thanks, makes sense. I reworked this down to the focused docs-only version you suggested: one new adapter table in docs/agent-portability.md, no duplicate Codex manifest, no CONTEXT.md, and no prompt behavior changes. I also updated it against current main so it reflects #7's Codex lifecycle hooks and the Kiro adapter.

PR #15, opened 2026-06-12
feat: add opencode plugin compatibility

gailingmic:
Add .opencode/ directory with plugin, commands, and skill definitions. Update hooks and statusline scripts to detect opencode flag path (~/.config/opencode/.ponytail-active) alongside existing Claude Code and Codex paths.

### DIFF
--- a/.opencode-plugin/plugin.json
+++ b/.opencode-plugin/plugin.json
@@ -1,0 +1,15 @@
+{
+  "name": "ponytail",
+  "version": "4.1.0",
+  "description": "Lazy senior dev mode. Forces the simplest, shortest solution that actually works: YAGNI, stdlib first, no unrequested abstractions.",
+  "author": {
+    "name": "Dietrich Gebert",
+    "url": "https://github.com/DietrichGebert"
+  },
+  "compatibility": ["claude-code", "opencode", "codex"],
+  "opencode": {
+    "plugin": ".opencode/plugins/ponytail.js",
+    "skills": ".opencode/skills",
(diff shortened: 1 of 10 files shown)


DietrichGebert:
Thanks for kicking off OpenCode support, @gailingmic. It prompted a proper dig into the OpenCode plugin API.

One blocker worth flagging: this plugin is built on hooks that don't exist in the current OpenCode API. `session.created` and `tui.prompt.append` aren't real OpenCode hooks, and there's no `output.systemPrompt` to mutate in a session hook, so as written it would load but silently no-op (nothing injected, no mode tracking).

The real mechanism (verified against OpenCode 1.17.4 source plus a live run):
- Inject via the `experimental.chat.system.transform` hook (mutate `output.system: string[]`).
- Server plugins `export default async (input) => hooks`.
- Plugins are discovered through the `opencode.json` `"plugin"` array, not a `.opencode-plugin/plugin.json` or a `package.json` block.

I've opened #16 with a version rebuilt on that API. It also reuses the shared instruction builder (`hooks/ponytail-instructions.js`) instead of copying SKILL.md and the filter/fallback logic, so there's one source of truth. End-to-end tested on 1.17.4: the hook fires and the ruleset reaches the model.

Credit to you for the idea and the nudge to support OpenCode 🙏

DietrichGebert:
Closing in favor of #16, which is rebuilt on the real OpenCode plugin API and verified end-to-end. Thanks again @gailingmic for raising OpenCode support.

PR #16, opened 2026-06-13
feat: add OpenCode adapter

DietrichGebert:
Adds OpenCode ([sst/opencode](https://github.com/sst/opencode)) as a host adapter, in keeping with the repo's agent-portability model.

## What it does
- `.opencode/plugins/ponytail.mjs`: server plugin. Injects the ponytail ruleset into every chat via the **`experimental.chat.system.transform`** hook at the active intensity, and persists `/ponytail <level>` switches via `command.execute.before`.
- Reuses the shared instruction builder (`hooks/ponytail-instructions.js`): **no copied SKILL.md, no duplicated filter/fallback logic, no edits to the shared hooks**. Stays within the "keep adapters thin" rule in `docs/agent-portability.md`.
- `.opencode/command/{ponytail,ponytail-review}.md`: slash commands.
- `tests/opencode-plugin.test.js`: smoke test (inject-at-mode, switch-follows, off-suppresses, ignore-other-commands).
- Example `opencode.json`, plus README / agent-portability / help-card docs.

## Verified against OpenCode 1.17.4
- Plugin loads cleanly (`opencode debug config`, no errors).
- Commands load into resolved config.
- The `experimental.chat.system.transform` hook fires and the injected system prompt reaches the model: Claude Sonnet 4.6 quoted the injected `PONYTAIL MODE ACTIVE` header; with mode `off` it correctly saw nothing.

## Relationship to #15
Supersedes #15. That PR's plugin is built on OpenCode hooks that don't exist in the current API (`session.created`, `tui.prompt.append`) and mutates a non-existent `output.systemPrompt`, so it would silently no-op. The real injection point is `experimental.chat.system.transform`. This is rebuilt on the verified API and slimmed to reuse the shared builder (71 LOC of adapter vs 319). Thanks to @gailingmic for raising OpenCode support.

### DIFF
--- a/.opencode/command/ponytail-review.md
+++ b/.opencode/command/ponytail-review.md
@@ -1,0 +1,5 @@
+---
+description: Review changes for over-engineering — what can be deleted
+---
+
+Review the current code changes for over-engineering only — not correctness. One line per finding: L<line>: <tag> <what to cut>. <replacement>. Tags: delete (dead code/speculative feature), stdlib (reinvented standard library), native (dependency doing what the platform does), yagni (abstraction with one implementation), shrink (same logic, fewer lines). End with the net lines removable. If nothing to cut: 'Lean already. Ship.'
--- a/.opencode/command/ponytail.md
+++ b/.opencode/command/ponytail.md
@@ -1,0 +1,5 @@
+---
+description: Switch ponytail intensity level (lite/full/ultra/off)
+---
+
+Switch to ponytail $ARGUMENTS mode. If no level specified, use full. Lazy senior dev mode — before any code: does it need to exist at all (YAGNI)? Does the standard library do it? A native platform feature? Can it be one line? Build the minimum that works. No unrequested abstractions, no avoidable dependencies, no boilerplate. Mark intentional simplifications with a ponytail: comment.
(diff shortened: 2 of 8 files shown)


vasilvestre:
I'm not sure about how to install this, I should download the repo locally ? Add it to my .config/opencode/ ?
Sorry 😓

JulienBernard:
Same difficulty than @vasilvestre , i try everything but I do not understand how to setup Ponytail under my Opencode configuration.

1 - Clone this repo inside `~/.config/opencode/tools/ponytail`
2 - Take `ponytail.mjs` from it and copy it to `~/.config/opencode/plugins/ponytail.mjs`
3 - Update `~/.config/opencode/opencode.json` to add it:
```
  "plugin": [
    "./plugins/ponytail.mjs"
  ]
```
4 - Start `opencode`
5 - No commands `/ponytail` available, no Ponytail skills visibles when I list them with `/skills`

The only working way is to clone the repo and start Opencode from it, but what the point if it's no available under our projects folders?

Edit: tried with the full path of ponytail.mjs into plugin, same issues, nothing happens.

Final Edit: fix PR incoming => https://github.com/DietrichGebert/ponytail/pull/143

PR #21, opened 2026-06-14
Update Windows hook environment variable syntax for PowerShell

AryanYadav215:
This updates the commandWindows paths to use PowerShell's $env: syntax instead of cmd's %VAR% syntax, which fixes the hook failures on Windows

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -7,7 +7,7 @@
           {
             "type": "command",
             "command": "node \"${CLAUDE_PLUGIN_ROOT}/hooks/ponytail-activate.js\"",
-            "commandWindows": "node \"%CLAUDE_PLUGIN_ROOT%\\hooks\\ponytail-activate.js\"",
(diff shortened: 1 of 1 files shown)


DietrichGebert:
Thanks so much for digging into this, @AryanYadav215 — you nailed the root cause. cmd's `%VAR%` not expanding under PowerShell was exactly it. 🙌

We ended up going with #26 for the fix, which landed the same insight with a slightly smaller footprint: it keeps the existing `node` + `.js` invocation and just swaps the `commandWindows` variable syntax over to `$env:`, plus adds a regression test so this can't quietly come back.

The only thing that held this one back: it repointed the commands at `.ps1` files (`ponytail-activate.ps1`) that don't ship in the repo yet, and the mode-tracker hook ended up aimed at the activate script — so the hooks would've still failed on Windows.

Closing as superseded, but genuinely grateful for the report and the diagnosis — that's what made #26 possible. Hope to see more PRs from you! 🙏

PR #24, opened 2026-06-14
feat: add ponytail-opencode-plugin standalone npm package

linhlban150612:
## Summary

Standalone npm-publishable OpenCode plugin for Ponytail lazy-senior-dev mode.

## What

- TypeScript + Bun build, modeled after `caveman-opencode-plugin`
- 3 hooks: `config`, `experimental.chat.system.transform`, `command.execute.before`
- Commands: `/ponytail <off|lite|full|ultra>`, `/ponytail-review`
- Config: `ponytail.json` (project) → `~/.config/opencode/ponytail.json` (global) → defaults
- Per-session state via in-memory Map
- 12 tests, all passing
- Zero runtime deps, peerDep on `@opencode-ai/plugin >=1.14.21`

## Files

```
ponytail-opencode-plugin/
├── package.json          tsconfig.json         build.ts
├── ponytail.json.example .gitignore            bun.lock
└── src/
    ├── index.ts          config.ts             state.ts
    ├── skills/ponytail.ts
    ├── commands/mode.ts  commands/review.ts
    └── __tests__/{config,state,commands}.test.ts
```

## Verification

- `tsc --noEmit` ✅
- `bun run build.ts` → dist/index.js ✅
- `bun test` → 12/12 pass ✅
- Companion for `@tarquinen/opencode-dcp` (context compression)


DietrichGebert (changes requested):
Thanks for this — the TypeScript, tests, and build setup are all solid work. The blocker is architectural fit rather than code quality: the package duplicates the one thing this repo most deliberately centralizes (the rule text), and it's already drifting from the canonical source.

**1. The ruleset is hand-copied — `src/skills/ponytail.ts` lines 3-80.**
The repo keeps a single source of truth for the rules (`AGENTS.md` + `skills/ponytail/SKILL.md`, with the shared builder in `hooks/ponytail-instructions.js`), and `scripts/check-rule-copies.js` guards the markdown copies against drift. This TS template literal is an unguarded copy, so it will silently fall out of sync the next time the rules change.

**2. It has already drifted.** `getReviewInstructions()` defines `yagni/bloat/nit/q`, but the canonical `skills/ponytail-review/SKILL.md` uses `delete/stdlib/native/yagni/shrink` — so this package would ship a different `/ponytail-review` than every other adapter. (Minor wording drift in the rule body too, e.g. "question the complex request" vs the canonical "question it".)

**3. `filterForMode` is reinvented.** The repo already exports `filterSkillBodyForMode` from `hooks/ponytail-instructions.js` (the pi extension reuses it).

**4. It forks the existing OpenCode adapter.** `.opencode/plugins/ponytail.mjs` (65 lines) already does `config` + `experimental.chat.system.transform` + `command.execute.before`, reusing the shared instruction builder. This PR reimplements all of that with its own config/state/build, creating a parallel maintenance surface.

The need is legit — a publishable `npm install`-able package so OpenCode users don't have to run from a checkout of this repo. A direction that keeps one source of truth:

- Have `build.ts` inline the rule text from `AGENTS.md` (or `skills/ponytail/SKILL.md`) at build time instead of hand-copying it into `src/skills/ponytail.ts`.
- Reuse `getPonytailInstructions` / `filterSkillBodyForMode` from `hooks/` rather than reimplementing them.
- Make the review command use the canonical `delete/stdlib/native/yagni/shrink` tags.
- Or, more simply: package the existing `.mjs` + `hooks/` as the npm module.

Happy to pair on the build-time-inlining approach if that helps. Thanks again for the contribution!

DietrichGebert:
Thanks a lot for this, and for caring about the OpenCode experience. I'm going to close it for now. Ponytail keeps a single source of truth in skills/ponytail/SKILL.md plus the shared hooks/, and the OpenCode integration we already ship reuses those directly, so everything stays in sync automatically. A standalone package that carries its own copy of the ruleset would need re-syncing every time the rules change, which is most weeks lately. If you ever want to revisit an approach that imports the shared ruleset instead of bundling a copy, I'd genuinely welcome it. Really appreciate the effort here.

PR #31, opened 2026-06-14
feat(benchmarks): add correctness assertion

zamal-db:
I was using the benchmark to compare arms but realized there's no signal for whether the generated code actually works - just LOC. Added a correctness assertion so broken outputs fail instead of silently scoring well.

## What

`correctness.js` is a promptfoo assertion (same pattern as `loc.js`) that extracts code from fenced blocks and runs per-task functional checks:

| Task | Method | Checks |
|------|--------|--------|
| Email validator | Spawns Python, calls the function | accept valid, reject 3 invalid inputs |
| Debounce | Spawns Node, runs with timers | no immediate fire, fires after delay |
| CSV sum | Spawns Python with a test CSV | correct total (351.0) |
| React countdown | Structural (no bundler needed) | useState + useEffect + decrement |
| FastAPI rate limit | Structural | limit logic + framework usage |

Python/JS tasks are execution-tested (spawn + assert). React/FastAPI use structural checks because running them needs a bundler/server - lazy, not negligent.

## How it fits

- Same file convention as `loc.js`: one assertion file, one metric (`correct`).
- Wired into `promptfooconfig.yaml` alongside the existing LOC metric.
- Zero new dependencies. Uses `child_process.execSync` + temp files.
- `loc.js` is a measurement (always passes). `correctness.js` is a gate (wrong answer = fail).

## Test verification (RED -> GREEN)

12 tests in `tests/correctness.test.js` (node:test, no frameworks):

```
node --test tests/correctness.test.js

# pass 12, fail 0
```

RED cases (bad outputs fed through the assertion):
- always-true email validator -> FAIL
- immediate-call debounce -> FAIL
- wrong CSV sum -> FAIL
- static div without state -> FAIL
- endpoint without limiting -> FAIL

GREEN cases (correct implementations):
- one-liner email check -> PASS
- setTimeout-based debounce -> PASS
- pandas one-liner -> PASS
- useState+useEffect countdown -> PASS
- FastAPI with HTTPException(429) -> PASS

Full suite unaffected:

```
node --test tests/*.test.js        # 19 pass, 0 fail
node scripts/check-rule-copies.js  # Rule copies match
```

### DIFF
--- a/benchmarks/correctness.js
+++ b/benchmarks/correctness.js
@@ -1,0 +1,261 @@
+// Functional correctness assertion: runs generated code against lightweight test
+// cases per task. Proves "less code" is not "broken code". Spawns python/node
+// with the extracted code + appended assertions; returns pass/fail + score.
+//
+// Metric: `correct` (1 = all checks pass, 0 = at least one fails).
+// Unlike loc.js (measurement-only), this one is a gate — a wrong answer is a
+// wrong answer regardless of how few lines produced it.
+
+const { execSync } = require('child_process');
+const fs = require('fs');
+const os = require('os');
+const path = require('path');
+
+// Extract fenced code blocks, tagged by language.
+function extractBlocks(text) {
+  const matches = [...text.matchAll(/```(\w*)\n([\s\S]*?)```/g)];
+  return matches.map((m) => ({ lang: (m[1] || '').toLowerCase(), code: m[2] }));
+}
+
+// Identify which task we're evaluating from vars.task.
+function identifyTask(task) {
+  const t = task.toLowerCase();
+  if (t.includes('email') && t.includes('valid')) return 'email';
+  if (t.includes('debounce')) return 'debounce';
+  if (t.includes('csv') && t.includes('sum')) return 'csv';
+  if (t.includes('countdown') && t.includes('react')) return 'countdown';
+  if (t.includes('rate limit') || t.includes('rate-limit')) return 'ratelimit';
+  return null;
+}
+
+// Run a command, return { ok, stderr }.
+function exec(cmd, opts = {}) {
+  try {
+    execSync(cmd, { timeout: 10_000, encoding: 'utf8', stdio: 'pipe', ...opts });
+    return { ok: true, stderr: '' };
+  } catch (e) {
+    return { ok: false, stderr: (e.stderr || e.message || '').slice(0, 500) };
+  }
+}
+
+// Write content to a temp file, return the path.
+function tmpFile(ext, content) {
+  const p = path.join(os.tmpdir(), `ponytail-bench-${Date.now()}-${Math.random().toString(36).slice(2)}${ext}`);
+  fs.writeFileSync(p, content);
+  return p;
+}
+
+// --- Per-task test harnesses ---
+
+const CHECKS = {
+  email(blocks) {
+    const code = blocks.find((b) => b.lang === 'python' || b.lang === 'py' || (!b.lang && b.code.includes('def ')));
+    if (!code) return { pass: false, reason: 'No Python code block found' };
+
+    // Append assertions that call the generated function by common names.
+    const harness = `
+${code.code}
+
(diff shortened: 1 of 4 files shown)


benchmarks/correctness.js
```diff
@@ -0,0 +1,261 @@
...
+output = sys.stdout.getvalue()
+sys.stdout = _stdout
+
+# Check output contains 351 (100.5 + 200.0 + 50.5)
+if '351' in output:
+    print("PASS")
+else:
+    # Try running it differently: maybe it defines a function
+    print("FAIL: output was: " + repr(output[:200]))
+    sys.exit(1)
```
DietrichGebert: This is the substring issue. `'351' in output` is also true for 13510, so a wrong sum can pass (I checked, `print(13510)` returns pass). This compares the actual number instead, and drops the dead retry comment while we're here.

```suggestion
# Pass only if the summed value is actually 351.0 (100.5 + 200.0 + 50.5),
# not just because the digits "351" appear somewhere (13510 would slip through).
vals = []
for tok in output.split():
    try:
        vals.append(float(tok.strip('$%,')))
    except ValueError:
        pass
if any(abs(v - 351.0) < 0.01 for v in vals):
    print("PASS")
else:
    print("FAIL: output was: " + repr(output[:200]))
    sys.exit(1)
```

benchmarks/correctness.js
```diff
@@ -0,0 +1,261 @@
...
+    if (!hasState) failures.push('no state management (useState/useReducer)');
+    if (!hasEffect) failures.push('no timer setup (useEffect/setInterval/setTimeout)');
+    if (!hasDecrement) failures.push('no countdown decrement logic');
+
+    if (failures.length === 0) return { pass: true, reason: 'Countdown has required structure' };
+    return { pass: false, reason: 'Missing: ' + failures.join(', ') };
+  },
+
+  ratelimit(blocks) {
+    const code = blocks.find((b) => b.lang === 'python' || b.lang === 'py' || (!b.lang && b.code.includes('rate') || b.code.includes('limit')));
```
DietrichGebert: Small precedence bug: `&&` binds tighter than `||`, so `|| b.code.includes('limit')` sits outside the `!b.lang` guard and matches any block containing "limit", whatever its language. The extra parentheses keep it scoped to untagged blocks.

```suggestion
    const code = blocks.find((b) => b.lang === 'python' || b.lang === 'py' || (!b.lang && (b.code.includes('rate') || b.code.includes('limit'))));
```

DietrichGebert (commented):
This is a great addition, thank you. The benchmark really was missing this. Measuring only LOC meant a short but broken answer could look like a win, and now it can't. Nice that you matched the loc.js pattern and added the tests.

I ran it here (python 3.13 + pandas 2.3.3, node 23) and the whole suite is green, 23 pass.

Two small things to fix before merge, I left both as suggestions you can apply directly:

1. The CSV check uses `'351' in output`, which is a substring match, so a wrong answer like 13510 passes because it contains "351". I tested it and `print(13510)` returns pass. Comparing the actual number fixes it.
2. A precedence bug in the ratelimit block finder, the `|| b.code.includes('limit')` ends up outside the `!b.lang` guard. One pair of parentheses scopes it right.

Not blockers, just nice to have: the React and FastAPI checks are keyword-based (you flagged this), so maybe a line in the README saying so, plus a note that the benchmark now needs python, pandas and node installed to run. Thanks again, this is a real improvement.

zamal-db:
thanks for the review! both fixed:

1. csv check now uses a regex with lookaround (`(?<!\d)351(?:\.0)?(?!\d)`) so 13510 no longer sneaks through. added a regression test for it too.
2. added parens around the `||` in the ratelimit block finder so it stays inside the `!b.lang` guard.

also added a note in the README that React/FastAPI checks are structural-only, plus a prerequisites line (python 3, pandas, node 18+).

13/13 tests green. should be good to merge now.

PR #47, opened 2026-06-15
Adding support for Copilot Marketplace plugin

maxfelker:
Adds official GitHub Copilot CLI plugin support for Ponytail — a thin adapter that points Copilot at the existing `commands/`, `skills/`, and hook runtime. No core rewrite, no forked logic.

## What's in it

- **Plugin + marketplace manifests** — `.github/plugin/plugin.json` and `.github/plugin/marketplace.json`, pointing at the repo's existing `commands/` and `skills/` directories.
- **One runtime branch, not a fork** — `hooks/ponytail-runtime.js` detects `COPILOT_PLUGIN_DATA`, routes state and `writeHookOutput` per host; Copilot reuses the shared `ponytail-activate.js` / `ponytail-mode-tracker.js`. Copilot and Codex detection are mutually exclusive.
- **Docs** — README and `docs/agent-portability.md` cover full plugin install vs. instruction-only fallback.

## Acceptance criteria

- [x] `.github/plugin/plugin.json` declares metadata + `commands`/`skills`/`hooks` component paths
- [x] `.github/plugin/marketplace.json` supports `copilot plugin marketplace add` → `install ponytail@ponytail`
- [x] Copilot hooks added without breaking existing Claude hook config
- [x] Copilot hook runtime stores state in `COPILOT_PLUGIN_DATA`, never `~/.claude` (proven in `tests/hooks.test.js`)
- [x] `/ponytail`, `lite`, `full`, `ultra`, `off` work in Copilot CLI
- [x] `/ponytail-review`, `/ponytail-audit`, `/ponytail-debt` available in Copilot CLI
- [x] `skills/` remain the single source of truth
- [x] `AGENTS.md` and `.github/copilot-instructions.md` remain fallback instruction-only paths
- [x] README documents both full plugin install and fallback mode
- [x] `docs/agent-portability.md` lists Copilot CLI as plugin-supported
- [x] JSON manifests parse cleanly
- [x] Existing Claude / Codex / OpenCode / Gemini support not regressed

## Review follow-ups (all addressed)

- [x] Version drift fixed — `.github/plugin/plugin.json` aligned to `4.4.0`
- [x] Version-parity test extended to all four versioned manifests (`tests/gemini-extension.test.js`)
- [x] `isCodex` made mutually exclusive with `isCopilot` (means Codex-only, no rename needed)
- [x] `/ponytail-debt` covered by a Copilot smoke test (`tests/copilot-plugin.test.js`)
[...]

PR #48, opened 2026-06-15
docs: add Copilot CLI plugin installation

jay-tau:
## Summary
- add explicit GitHub Copilot CLI plugin marketplace commands
- mention that ponytail skills appear under `/skills` after starting a new session
- update the portability matrix and help skill so Copilot CLI is not documented as instruction-only

## Validation
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,16 @@ The most effort ponytail will ever ask of you:
 /plugin install ponytail@ponytail
 ```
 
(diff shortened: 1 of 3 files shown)


jay-tau:
Closing as redundant: #47 already adds full GitHub Copilot CLI plugin support and the install docs.

PR #50, opened 2026-06-15
fix: use python3 for correctness checks and add CI

christophermayfield:
## Summary
- Fix the correctness benchmark harness to probe `python3` before `python` (macOS and many CI images ship only `python3`, which caused email and CSV checks to fail locally)
- Add a root `npm test` script that runs the repo and pi-extension test suites
- Add a GitHub Actions workflow that runs rule-copy checks and the full test suite on every push/PR

## Test plan
- [x] `npm test` passes locally (43 tests)
- [x] `node scripts/check-rule-copies.js` passes
- [ ] CI workflow passes on this PR

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,29 @@
+name: test
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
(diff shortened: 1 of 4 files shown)


DietrichGebert (approved):
This is great, thank you. Three things I'd been meaning to do and you knocked out all of them: the python3 probe (exactly why the correctness tests "failed" for people on mac), an actual npm test, and CI, which the repo flat out didn't have. Ran npm test here and it's green, and nice touch marking the probe with a ponytail: comment. Merging.

PR #57, opened 2026-06-15
fix: hooks degrade gracefully when node is not on PATH

### LINKED ISSUE #51: UserPromptSubmit hook errors on every prompt when node isn't on a global PATH (undeclared dependency)
The `UserPromptSubmit` and `SessionStart` hooks invoke `node` directly:

```json
"command": "node \"${CLAUDE_PLUGIN_ROOT}/hooks/ponytail-mode-tracker.js\""
```

Claude Code runs hooks via a non-interactive `/bin/sh`. On systems where `node` isn't on that shell's `PATH`, every prompt submit fails with:

```
UserPromptSubmit hook error
/bin/sh: node: command not found
```

This happens on Nix / nix-darwin setups (and others) where node is intentionally scoped to a dev shell rather than installed globally. It's reproducible on a fresh install: `node` is never declared as a prerequisite — not in `package.json` (no `engines`), not in the README's Install section, not in AGENTS.md. The only `node` reference in the docs is a contributor script (`node scripts/check-rule-copies.js`).

Two things worth fixing, independently:

1. **The mode-tracker hook is non-essential but fails loudly.** A `UserPromptSubmit` tracker erroring on every single prompt is noisy for something that's just bookkeeping. It'd be better to degrade silently when node is unavailable (e.g. wrap in a shell guard like `command -v node >/dev/null && node ... || exit 0`, or detect a node path and no-op if absent). The core skills still work, so this is pure noise.

2. **node is an undeclared dependency.** If a global node is genuinely required, the Install section should say so. For a plugin whose entire philosophy is "don't add dependencies, use what's already there," a silent assumption of a global node runtime feels worth calling out. 🙂

Happy to test a patch on a node-less PATH if useful.

**Environment:** ponytail 4.4.0, Claude Code, macOS (nix-darwin), node only present inside `nix shell`.

DietrichGebert:
Fixes #51.

Claude Code runs hooks through a non-interactive `/bin/sh`. On setups where `node` isn't on that shell's PATH (Nix/nix-darwin, nvm, fnm), the `SessionStart` and `UserPromptSubmit` hooks failed with `/bin/sh: node: command not found`, **on every prompt**.

- Guard each hook command: `command -v node >/dev/null 2>&1 && node "..." || exit 0` (and a `Get-Command node` guard for the PowerShell `commandWindows`). Node-absent now exits 0 silently instead of erroring every prompt.
- The slash-command skills are unaffected (Claude Code discovers `skills/` directly); only the always-on activation + mode tracking need node.
- Document the node requirement in the README install section, with a Nix/nvm note.

Verified: hooks.json parses, the guard runs node when present and exits 0 silently when absent, full suite 33/33, rule check green. @githappens offered to test on a node-less PATH.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,8 @@ Lazy, not negligent: trust-boundary validation, data-loss handling, security, an
 
 The most effort ponytail will ever ask of you:
 
+The Claude Code and Codex plugins run two tiny Node.js lifecycle hooks, so `node` needs to be on your PATH (note for Nix/nvm users: it must be on the non-interactive shell's PATH). If it isn't, the skills still work, the always-on activation just stays quiet instead of erroring on every prompt.
+
 ### Claude Code
 
 ```
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -6,8 +6,8 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node \"${CLAUDE_PLUGIN_ROOT}/hooks/ponytail-activate.js\"",
-            "commandWindows": "node \"$env:CLAUDE_PLUGIN_ROOT\\hooks\\ponytail-activate.js\"",
+            "command": "command -v node >/dev/null 2>&1 && node \"${CLAUDE_PLUGIN_ROOT}/hooks/ponytail-activate.js\" || exit 0",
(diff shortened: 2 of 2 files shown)


githappens:
Confirmed fixed on my end — pulled `main` HEAD (`b41cb8d`) via a clean marketplace remove/re-add/reinstall, and the guarded hooks now exit 0 silently instead of erroring on every prompt. No more `/bin/sh: node: command not found` on my Nix/nix-darwin setup (node scoped to a dev shell, intentionally off the global PATH). Fast turnaround — thanks!

PR #63, opened 2026-06-15
benchmarks: add local model support and Node version note

### LINKED ISSUE #54: Local model benchmark support + Node.js version note
Hi,  I ran your benchmark against a local llama3.2 (3.2B via Ollama) and found two gaps worth addressing:

**1. promptfoo fails on Node 22.17.x** :  `npx promptfoo eval` exits with an engine error on Node >=22.0 but <22.22.0. This silently blocks a lot of users who upgraded Node recently. A one-line note in benchmarks/README.md would save them.

**2. No local/Ollama benchmark path** : the existing tooling requires an Anthropic API key. I wrote a small Python script (no promptfoo, no API key) that runs the same 5 tasks against any Ollama model and produces the same LOC + time table.

The results are interesting: ponytail reduces LOC by only ~5% on llama3.2 vs 80–94% on Claude — the skill is calibrated for strong instruction-following models, and a small local model doesn't reliably follow the decision ladder. I think this is worth documenting honestly.

I have a branch ready with:
- `benchmarks/benchmark-local.py` : the Ollama runner
- `benchmarks/results/2026-06-15-llama3.2-local.md` : the llama3.2 results writeup
- `benchmarks/README.md` : updated Reproduce section with both paths

Happy to open a PR if this direction sounds useful to you.

mandavillivijay:
Closes #54

Three small additions to the benchmark suite, all motivated by actually running the benchmark:

## What's in this PR

**`benchmarks/benchmark-local.py`** — Ollama benchmark runner that needs no promptfoo and no API key. Drop-in replacement for the promptfoo path for anyone with a local model. Flags: `--model`, `--repeat`, `--ollama-url`.

**`benchmarks/results/2026-06-15-llama3.2-local.md`** — Results from running against llama3.2 (3.2B Q4_K_M) locally. The honest finding: ponytail reduces LOC by ~5% on llama3.2 vs 80–94% on Claude. The skill is calibrated for strong instruction-followers; a 3B model partially absorbs the decision ladder and then adds prose justifying its choices, paying the complexity cost without the minimalism benefit. Worth documenting rather than hiding.

**Code LOC**

| arm | email | debounce | csv-sum | countdown | rate-limit | TOTAL |
|---|--:|--:|--:|--:|--:|--:|
| baseline | 13 | 13 | 38 | 44 | 32 | **140** |
| caveman | 16 | 12 | 5 | 45 | 28 | **106** |
| ponytail | 18 | 21 | 7 | 38 | 49 | **133** |

**Time (seconds)**

| arm | email | debounce | csv-sum | countdown | rate-limit | TOTAL |
|---|--:|--:|--:|--:|--:|--:|
| baseline | 52.6 | 37.2 | 63.4 | 65.2 | 71.0 | **289.4** |
| caveman | 79.6 | 54.4 | 28.3 | 71.4 | 61.0 | **294.7** |
| ponytail | 99.7 | 71.0 | 25.4 | 74.8 | 97.3 | **368.2** |

On 3 of 5 tasks (email, debounce, rate-limit) ponytail produced more code than the no-skill baseline. Caveman outperformed ponytail on this model (−24% vs −5%) because its rules are simpler prose a small model can follow without deliberating.

**`benchmarks/README.md`** — Splits the Reproduce section into a Claude path and an Ollama path, and adds a note that promptfoo requires Node ≥22.22.0 (it silently blocks users on Node 22.17.x, which is a common version post the LTS upgrade).

## Test

```bash
ollama pull llama3.2
python benchmarks/benchmark-local.py --model llama3.2

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -4,11 +4,29 @@ Three arms (no skill, [caveman](https://github.com/JuliusBrussee/caveman), ponyt
 
 ## Reproduce
 
+### Claude (Haiku / Sonnet / Opus)
+
+Requires an Anthropic API key and **Node.js ≥ 22.22.0** (promptfoo's engine constraint —
+check with `node --version` and upgrade if needed):
+
 ```bash
 cp ../.env.example ../.env      # add your ANTHROPIC_API_KEY
 npx promptfoo@latest eval -c promptfooconfig.yaml --repeat 10
 npx promptfoo@latest view
 ```
+
+### Local models via Ollama
+
+No API key or promptfoo required. Runs against any model served by Ollama:
+
+```bash
+ollama pull llama3.2          # or any other model
+python benchmarks/benchmark-local.py --model llama3.2 --repeat 3
+```
+
+See `benchmarks/results/2026-06-15-llama3.2-local.md` for what to expect: the skill works
+well on instruction-following models (Claude-class) but transfers poorly to small local
+models where the multi-step decision ladder isn't reliably followed.
 
 Tasks: email validator, JS debounce, CSV sum, React countdown, FastAPI rate-limit (see `promptfooconfig.yaml`). Single-shot completions, default temperature.
 
--- a/benchmarks/benchmark-local.py
+++ b/benchmarks/benchmark-local.py
@@ -1,0 +1,155 @@
+"""
+Ponytail local benchmark — runs the same 5 tasks against any Ollama model.
+No promptfoo required. Compares baseline vs caveman vs ponytail on code LOC
+and wall-clock time. Results are printed as a table and saved to a JSON file.
+
+Usage:
+    python benchmarks/benchmark-local.py
+    python benchmarks/benchmark-local.py --model llama3.2 --repeat 3
+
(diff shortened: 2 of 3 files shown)


DietrichGebert:
Merged, thank you! I reproduced it locally end to end (Ollama + llama3.2) to confirm it runs. Two things the live run surfaced, both fixed in the follow-up #67:

1. The LOC counter (and its `loc.js` twin) only counted code inside fenced, triple-backtick blocks. llama3.2 often returns bare, unfenced code, which scored 0 and quietly deflated whichever arm skipped the fences.

2. The summary's final `print` used a Unicode arrow that crashes the script with a non-zero exit on Windows (cp1252), even after all the work has succeeded. Now ASCII.

On the numbers: once the counter is fixed and I run it a few times (n=3 and n=5), the LOC effect on llama3.2 turns out to be inside the noise floor. Per-run totals swing from 17% below baseline to 50% above, so there is no stable reduction either way. That actually backs up your core conclusion that ponytail does not transfer to a 3B local model. I refreshed your results file with the n=5 data and a note about the counter fix. Great addition, the local runner fills a real gap.

PR #69, opened 2026-06-15
Add missing ponytail-help command files

Injulminuget:
## Summary

This PR adds the missing `/ponytail-help` command surfaces so the README command list matches the actual command files.

## Changes

- Added `commands/ponytail-help.toml`
- Added `.opencode/command/ponytail-help.md`
- Updated the Copilot command smoke test to require `ponytail-help.toml`

## Tests

- `node tests\copilot-plugin.test.js`
- `node tests\gemini-extension.test.js`
- `node tests\opencode-plugin.test.js`
- `node scripts\check-rule-copies.js`

Note: `npm.cmd test` could not complete in this local sandbox because Node's test runner failed to spawn test-file subprocesses with `EPERM`; the targeted tests above passed when run directly.

### DIFF
--- a/.opencode/command/ponytail-help.md
+++ b/.opencode/command/ponytail-help.md
@@ -1,0 +1,5 @@
+---
+description: Show quick reference for Ponytail commands
+---
+
+Display the Ponytail quick reference card. One-shot display only: do not change mode, write flag files, or persist anything. Include levels, command list, deactivation, default mode configuration, update flow, and full docs link.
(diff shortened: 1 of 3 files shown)


DietrichGebert:
Thanks for catching this and sending a fix! Closing as a duplicate of #62, which just merged and ships the same two ponytail-help adapter files plus a parity test that guards against this whole class of gap going forward. Genuinely the same good catch, I appreciate you taking the time. The command is on main now.

PR #77, opened 2026-06-15
fix: emit per-turn reminder in UserPromptSubmit to prevent drift after compaction

vanceingalls:
## Problem

`SessionStart` injects the full ponytail ruleset once. When the context window fills and compacts, that injection disappears. The `UserPromptSubmit` hook only fires context on `/ponytail` commands — silent the rest of the time. Result: model drifts back to over-building mid-session with no reminder.

## Fix

`ponytail-mode-tracker.js` now emits a brief `PONYTAIL MODE ACTIVE — level: {mode}` reminder on every turn (when active), matching the pattern already used by tools like Caveman. This keeps the active mode in context across compaction boundaries.

Changes:
- `hooks/ponytail-runtime.js`: export `getMode()` — reads current mode from the flag file
- `hooks/ponytail-mode-tracker.js`: import `getMode()`, add early `return` on mode-change paths, emit per-turn reminder at the end of the normal (non-command) path

The reminder is one line, zero overhead when ponytail is off.

### DIFF
--- a/hooks/ponytail-mode-tracker.js
+++ b/hooks/ponytail-mode-tracker.js
@@ -3,14 +3,14 @@
 // Inspects user input for /ponytail commands and writes mode to flag file
 
 const { getDefaultMode } = require('./ponytail-config');
-const { clearMode, setMode, writeHookOutput } = require('./ponytail-runtime');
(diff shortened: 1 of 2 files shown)


anupddas (commented):
👍

PR #79, opened 2026-06-15
Add an opt-in readability guard (off by default)

VforVitorio:
Ponytail already says "lazy means efficient, not careless" and keeps input validation, data-loss handling, security and accessibility off the chopping block. One thing that isn't on that list: readability.

The whole point is to lean on being a senior and write little code — but "little" should never quietly cost the things a senior actually protects: sound architecture, readability, scalability, maintainability, modularity. A one-line lambda that swallows two well-named helpers and now takes a minute to decode isn't the senior move, it's a junior flexing. "Boring over clever" hints at this, but it's a soft preference, not a guard.

This adds it as an opt-in, off by default, so nothing changes for anyone who doesn't want it:

- `PONYTAIL_GUARD_READABILITY=1`, or `"guardReadability": true` in `~/.config/ponytail/config.json` (`%APPDATA%\ponytail\config.json` on Windows)
- When on, the hook appends one clause to the active level: never buy fewer lines with worse architecture, readability, scalability, maintainability or modularity; when short and clear pull apart, clear wins — and name the corner you cut in one line.

A few implementation notes:

- It rides the existing config resolver (`ponytail-config.js`, same place as `defaultMode`): env first, then config, default off.
- It only touches the hook-injected ruleset (`ponytail-instructions.js`). The canonical rule files (`AGENTS.md` and the `.cursor` / `.windsurf` / `.clinerules` / `.kiro` / copilot copies) are untouched, so `check-rule-copies.js` stays green.
- `writeDefaultMode` now merges the config file instead of overwriting it, so `/ponytail <level>` no longer wipes the flag.
- Added `tests/readability-guard.test.js` (off → clause absent; env/config on → clause present and names the dimensions). `npm test` passes locally; the one CSV correctness check needs `pandas` on the machine, unrelated to this change.

Happy to reword the clause, or gate it behind a level instead of a flag, if you'd prefer — this felt like the least invasive shape.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,12 @@ Active every session, with a handful of commands (see [Commands](#commands)). `/
 
 Set the level for every new session with the `PONYTAIL_DEFAULT_MODE` env var (`lite`/`full`/`ultra`/`off`), or a `defaultMode` field in `~/.config/ponytail/config.json` (`%APPDATA%\ponytail\config.json` on Windows). The default is `full`.
 
-**Readability guard (opt-in, off by default).** Writing little is the senior move, not the goal. Turn this on and ponytail adds one rule to the active level: never buy fewer lines with worse architecture, readability, scalability, maintainability, or modularity — when short and clear pull apart, clear wins. Enable with `PONYTAIL_GUARD_READABILITY=1` or `"guardReadability": true` in the same config file. Nothing changes unless you opt in.
+**Readability guard (opt-in, off by default).** Writing little is the senior move, not the goal. Turn this on and ponytail adds one rule to the active level: never buy fewer lines with worse architecture, readability, scalability, maintainability, or modularity — when short and clear pull apart, clear wins.
+
+- **Turn it on:** `PONYTAIL_GUARD_READABILITY=1` (env), or `"guardReadability": true` in `~/.config/ponytail/config.json` (`%APPDATA%\ponytail\config.json` on Windows).
+- **Turn it off:** drop the env var (or set it to `0`), or `"guardReadability": false`. It's off by default, so doing nothing keeps it off.
+- **Precedence:** the env var wins over the config file. The env var is read at session start; the config file is read each turn.
+- **What it changes:** when on, the agent gets one extra rule at the top of every turn — no slash command, no new dependency. Everyone who doesn't set it sees ponytail exactly as before.
 
 Cursor, Windsurf, Cline, GitHub Copilot (editor), Aider, Kiro: copy the matching rules file from this repo ([`.cursor/rules/`](.cursor/rules/), [`.windsurf/rules/`](.windsurf/rules/), [`.clinerules/`](.clinerules/), [`.github/copilot-instructions.md`](.github/copilot-instructions.md), [`AGENTS.md`](AGENTS.md), [`.kiro/steering/`](.kiro/steering/)).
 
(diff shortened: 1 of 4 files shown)


VforVitorio:
Small follow-up (87d2048): expanded the README so the on/off path is explicit — env vs config file, which one wins, that the env var is read at session start and the config each turn, and that it's off by default so doing nothing keeps it off. The first pass only said "off by default" and left disabling implicit. Docs only, no behaviour change.

VforVitorio:
Opened #112 to discuss the underlying question on its own — whether readability belongs on the never-simplify-away list at all, and default-on vs opt-in. This PR is just one possible shape for it.

Lakshya77089:
Pulled this down and ran it on Windows (node 20.20.0). `node --test tests/*.test.js` → **37/37 pass**, including the 9 guard assertions. Confirmed behavior matches the description: off by default (no clause), env and config both enable, env wins over config, explicit env `off` overrides config `on`, and the clause names all five dimensions. The `writeDefaultMode` merge so `/ponytail <level>` no longer wipes the flag is a nice catch. Canonical rule files untouched.

Unrelated note for Windows testers: `npm test` fails here because the `tests/*.test.js` glob in the script isn't expanded by cmd.exe and node 20 doesn't glob `--test` paths natively (node 21+ does). Expanding the glob passes clean — pre-existing in the script on `main`, not from this PR.

Implementation LGTM. The default-on vs opt-in question is #112's to settle.

DietrichGebert:
This is a thoughtful one, thank you, and you found a real failure mode. Models do over-collapse when pushed toward "one line," and you are right that "boring over clever" is a preference line, not a hard guard. Fair hit.

I am going to pass on the PR though, and the reason is the cure, not the problem.

Ponytail already gives you readability, but as a byproduct, not a goal. It comes from writing less plus "boring over clever." The risk in making it a named guard is the words you picked to protect: architecture, scalability, maintainability, modularity. Those are exactly the terms that push a model toward more structure and more abstraction, which is the over-engineering the skill exists to remove. So a guard meant to protect clarity would, in practice, invite the bloat I am trying to kill.

There is also a coverage gap. The flag only lives in the hooks, so it does nothing on the static-file integrations (Cursor, Windsurf, Cline, Kiro, Gemini, Antigravity). A reader on those gets no guard even with the flag set. Opt-in and universal cannot both be true here, since the static copies have no runtime config.

Where you genuinely landed something is the wording. "Fewer lines is never worth code someone has to decode" is good. That is the right shape: one line in the existing "boring over clever" rule, shipped to every copy, no flag, no config. But I will not add it on feel. It rides in the always-on ruleset every turn, so before it goes anywhere I want to benchmark it and confirm it does not quietly push code longer or move the correctness numbers. If it earns its place, in it goes.

So I will close this, with real thanks for the care and the tests. The instinct is right, and that one line might well survive the benchmark. I just do not add rules to the always-on prompt without measuring them first. I will close #112 as well and track the wording test on my side, with credit to you if it ships.

VforVitorio:
@DietrichGebert Fair call — the static-integration coverage gap is the part I hadn't weighed, and you're right that those four words pull toward the very structure the skill exists to cut. Thanks for the careful read and for taking it seriously. If that one line survives the benchmark, that's a great outcome either way. Cheers!

PR #78, opened 2026-06-15
add Hermes plugin for Ponytail with command support and documentation

salaamdev:
## Issues
Closes #125 
Closes #133

## Summary

Adds native [Hermes Agent](https://hermes-agent.nousresearch.com/) plugin support to Ponytail, installable with one command, injects the active ruleset before every LLM turn, and exposes all Ponytail commands and skills through Hermes's plugin system.

## Motivation

Hermes Agent is a gateway/agent framework with its own plugin API (hooks, skills, slash commands). This PR integrates Ponytail as a first-class Hermes plugin so Hermes users get the same always-on lazy-dev mode and command surface that Claude Code, Codex, OpenCode, Gemini, and pi users already have.

## What it does

**Install:**
```bash
hermes plugins install DietrichGebert/ponytail --enable
```

**Hooks registered:**
- `pre_llm_call` — injects the mode-filtered Ponytail ruleset (from `skills/ponytail/SKILL.md`) as system context before each LLM turn. Supports `lite`/`full`/`ultra` modes, config via `PONYTAIL_DEFAULT_MODE` env or `~/.config/ponytail/config.json`, and a `review` mode that loads the review skill.
- `pre_gateway_dispatch` — rewrites `/ponytail-review`, `/ponytail-audit`, `/ponytail-debt`, `/ponytail-help` gateway slash commands into agent prompts (with access-control delegation to Hermes's `_check_slash_access`).

**Slash commands:**
- `/ponytail [lite|full|ultra|off]` — get/set runtime mode (process-local)
- `/ponytail-review [target]` — review diff for over-engineering
- `/ponytail-audit [target]` — whole-repo audit
- `/ponytail-debt` — harvest `ponytail:` shortcuts into a ledger
- `/ponytail-help` — quick reference

**Skills:** All bundled skills (`ponytail`, `ponytail-review`, `ponytail-audit`, `ponytail-debt`, `ponytail-help`) registered under the `ponytail:` namespace.

## Files added/changed

| File | Change |
|------|--------|
| `__init__.py` | Plugin implementation (216 lines) — hooks, commands, skill registration, mode management, config resolution, gateway command rewriting |
| `plugin.yaml` | Plugin manifest — declares hooks, commands, and skills for `hermes plugins install` |
| `after-install.md` | Post-install instructions shown after `hermes plugins install` |
[...]

PR #80, opened 2026-06-15
Add Hermes Agent plugin

AtomicCactus:
## Summary

Adds a minimal Hermes Agent adapter for Ponytail.

The adapter reuses the existing `skills/` content and exposes Ponytail to Hermes as an explicit opt-in plugin. Hermes starts with Ponytail OFF by default. Users can turn it ON with `/ponytail on`, `/ponytail lite`, `/ponytail full`, or `/ponytail ultra`, and turn it off with `/ponytail off`.

The reason it's off by default is due to some users not always using their Hermes Agents for coding, so it would be wrong to have this plugin always enabled. It's also tricky to know when is the right time to automatically enable it, so users can exercise discretion and ask their agent to enable/disable the plugin automatically if they don't want to explicitly turn it on/off.

## Hermes Behavior

```bash
hermes plugins install DietrichGebert/ponytail --enable
```

After restart, Hermes loads the plugin but does not inject Ponytail until the user explicitly turns it on:

- `/ponytail on` --> defaults to `full` mode
- `/ponytail full`
- `/ponytail lite`
- `/ponytail ultra`
- `/ponytail off` --> off

When active, the plugin injects the selected Ponytail ruleset through Hermes' `pre_llm_call` hook.

## Validation

- Python smoke test for Hermes adapter registration and strict on/off behavior
- `PATH="$PWD/.venv/bin:$PATH" npm test`
  - Ponytail tests: 36 passed
  - Pi extension tests: 11 passed
- Installed locally with `hermes plugins install file://... --enable --force`
- Verified Hermes plugin discovery, command registration, skill registration, default-off behavior, strict slash-command mode injection, and no natural-language shutdown handling
- Ran it on this PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
 <p align="center">
   <img src="https://img.shields.io/github/stars/DietrichGebert/ponytail?style=flat-square&color=111111&label=stars" alt="Stars">
   <img src="https://img.shields.io/github/v/release/DietrichGebert/ponytail?style=flat-square&color=111111&label=release" alt="Release">
-  <img src="https://img.shields.io/badge/works%20with-13%20agents-111111?style=flat-square" alt="Works with 13 agents">
+  <img src="https://img.shields.io/badge/works%20with-14%20agents-111111?style=flat-square" alt="Works with 14 agents">
   <img src="https://img.shields.io/badge/license-MIT-111111?style=flat-square" alt="MIT license">
 </p>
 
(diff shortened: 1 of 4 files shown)


salaamdev:
There is already a hermes plugin pr open [#78](https://github.com/DietrichGebert/ponytail/pull/78)

this pr seems like a copy/reword of the already open hermes plugin pr

DietrichGebert:
Thanks for building this. #78 takes the same Hermes Python-plugin approach but more complete (mode persistence + fuller tests), so I'm consolidating on that one to avoid two parallel plugins. Closing this as superseded by #78, appreciate the contribution.

PR #82, opened 2026-06-15
feat: add Modern Web Guidance as the rung-3 lookup for web tasks

### LINKED ISSUE #81: feat: refer to Modern Web Guidance for native-platform lookups on web tasks
Rung 3 says "native platform feature covers it? Use it" - but on web tasks it never says *which* feature. That's exactly where models reach for a library, because their training data is full of legacy patterns. The README's `<input type="date">` win only happens if the agent remembers the platform has it. Often it doesn't.

[Modern Web Guidance](https://github.com/GoogleChrome/modern-web-guidance) (GoogleChrome, Apache-2.0) is a `SKILL.md` + offline CLI that returns curated native-platform patterns into context. It's the lookup layer for rung 3 on web work.

**Proposal:** when a web task turns on whether the platform covers it and the `modern-web` CLI is present, the agent looks the feature up - then runs the result through the ladder as usual.

Kept narrow, to stay true to ponytail's voice:

- **Lookup, not license.** MWG pushes the cutting edge (scrollytelling, WebGL shaders); ponytail keeps only the rung that holds. The note says so, so it doesn't read as "adopt new APIs."
- **Optional.** No dependency, install step, or hook. Absent CLI changes nothing - zero-config stays zero-config.
- **No benchmark claims.** Proving MWG helps is its own eval harness's job, out of scope here.

**Scope:** a ~10-line section in `skills/ponytail/SKILL.md` (runtime source, not byte-compared), a one-line "Pairs well with" in the README by the existing Caveman cross-ref, and a worked `examples/web-platform-lookup.md` mirroring `date-picker.md`. The compact body and its synced copies stay untouched, no new INVARIANT phrase, so `check-rule-copies.js` and `npm test` stay green.

liady:
Closes #81.

Adds a scoped, optional reference to Modern Web Guidance so the agent can look up native platform features on web work - the gap at rung 3 of the ladder ("use the native feature" never says *which* one).

Three additive changes, no compact-ruleset surgery:

- `skills/ponytail/SKILL.md` - a "Web tasks: rung 3 lookup" section after the ladder. Runtime source only, not a compact copy, so no six-file byte-sync needed.
- `README.md` - one "Pairs well with" line, matching the existing Caveman cross-reference.
- `examples/web-platform-lookup.md` - a `<dialog closedby>` vs. Radix before/after, in the `date-picker.md` style.

Framed as lookup, not license: MWG suggests, the ladder filters. Optional - an absent CLI changes nothing.

`node scripts/check-rule-copies.js` passes and `npm test` is green except `csv: correct pandas one-liner passes`, which is environmental (needs `pandas` installed locally, per the README's Development note) and fails identically on a clean checkout. The shared compact body and its synced copies are untouched, and no new INVARIANT phrase is added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,8 @@ With ponytail:
 
 More survivors in [examples/](examples/).
 
+> **Pairs well with** [Modern Web Guidance](https://github.com/GoogleChrome/modern-web-guidance) for web work: ponytail decides *whether* to lean on the platform, MWG is how the agent looks up *which* native feature does the job. See [examples/web-platform-lookup.md](examples/web-platform-lookup.md).
+
 ## Numbers
 
 Five everyday tasks (email validator, debounce, CSV sum, countdown timer, rate limiter), three models, three arms: no skill, the [caveman](https://github.com/JuliusBrussee/caveman) skill, and ponytail. Ten runs per cell, median reported.
--- a/examples/web-platform-lookup.md
+++ b/examples/web-platform-lookup.md
@@ -1,0 +1,64 @@
+# Web Platform Lookup
+
+**Task:** "Add a modal dialog that closes when you click the backdrop."
+
+Rung 3 of the ladder is "native platform feature covers it?" On web work the
+trap is that the agent forgets what the platform already does and reaches for a
+library. When ponytail has [Modern Web Guidance](https://github.com/GoogleChrome/modern-web-guidance)
+on hand, rung 3 gets a lookup: `modern-web search "modal dialog light dismiss"`.
+
+## Without Ponytail
+
+```bash
+npm install @radix-ui/react-dialog
+```
+
+```jsx
+import * as Dialog from "@radix-ui/react-dialog";
+
+export default function Modal({ open, onOpenChange, children }) {
+  return (
+    <Dialog.Root open={open} onOpenChange={onOpenChange}>
+      <Dialog.Portal>
+        <Dialog.Overlay className="overlay" />
+        <Dialog.Content className="content">
+          {children}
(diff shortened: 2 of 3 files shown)


anupddas (approved):
👍

DietrichGebert:
Reverted in #178, want to explain why since the idea was reasonable.

Two reasons. First, it puts a section about an external tool (the `modern-web` CLI) into the always-on SKILL.md, and that ruleset rides every turn for every user — most of whom won't have the CLI installed. So it's weight in the prompt that mostly can't be used, which is the kind of thing the ladder itself argues against. Second, it shipped without regenerating the `.openclaw` skill mirror, so it broke CI.

Appreciate the contribution, the always-on core just needs to stay lean and self-contained. If `modern-web` is worth surfacing, it fits better as an optional note in `examples/` than in the core skill.

liady:
Thanks, the broken CI was due to the branch not fully updated from recent upstream fixed it. 
I totally understand the ask to leave SKILL.md untouched. 

Opened a leaner, `examples/` only PR, as https://github.com/DietrichGebert/ponytail/pull/191 .

MWG is a natural companion to Ponytail's philosophy (especially for the web) so there might be value in adding at least a line / example about it.

I reworked it to live entirely in `examples/`, per your steer: the example now carries the lookup mechanic and the "optional, skip if not installed" framing, the README gets one "Pairs well with" line pointing at it, and `SKILL.md` is left byte-identical to main, so there is no always-on weight.

Happy to drop the README line too if it is still more surface than you want - the example stands alone. Thanks for taking the time to explain the revert.

Issue #65, opened 2026-06-15
Impact on model performance?


ForsakenHarmony:
There's a few of these, but none ever measure performance, it would be interesting to see the results on `SWE-Bench Pro` and `Terminal Bench 2.1` compared to stock.

web-max:
This is a really good question

DietrichGebert:
Fair ask, honest answer: I haven't run SWE-Bench Pro or Terminal Bench. The current benchmark measures code size, cost, latency, and a correctness gate on five everyday tasks, not task-success on a full agentic harness.

My prior for SWE-Bench: ponytail shouldn't move resolution rate much, because it explicitly refuses to simplify away correctness, validation, security, or anything requested. It trims over-engineering, not the fix. The delta I'd expect is in tokens/cost per solved instance. But that's a hypothesis, not a number, and I won't put an unmeasured number in the README. If anyone wants to wire ponytail in as a SWE-Bench arm, I'm very interested either way.

Related: v4.6.0 just shipped a local-model run showing the LOC win evaporates on a 3B model, so the effect is clearly model-dependent.

guplem:
I've seen many system prompts that reduce token usage, explain things clearly, and plan better. But I never see them tested against benchmarks, so how can we know they’re working?

I’ve so far refused to use any because I always think, "_If they were as good, they would be added to ClaudeCode, OpenCode, Codex, and so on_." 

So maybe they aren’t as good, but how much worse are they? If they’re 95% there and reduce usage by more than 80%, I would instantly switch!

But since there are no official benchmarks run, here I am, taking a look at the prompts to see if I can improve my system prompts a bit ^^'

Tobbe:
> I’ve so far refused to use any because I always think, "_If they were as good, they would be added to ClaudeCode, OpenCode, Codex, and so on_."

I tend to agree. But we also have to take into account that ClaudeCode and Codex both have an economic incentive to make you spend **more** tokens, not less. So from your list, I think OpenCode would be the most interested in implementing something like this natively if it really does improve things.

guplem:
> both have an economic incentive to make you spend more tokens

Yeah, but... don’t they charge whatever percentage they feel like based on usage? I’m not familiar with Codex, but with Claude Code you don’t really know how many tokens you can use per session. They increase the percentage used, and it varies by model and even from week to week. So they could simply change the token–percentage relationship while reducing server load and costs. I don’t know how they would handle API users, though

[...]

PR #83, opened 2026-06-16
benchmarks: fix correctness gate + robustness audit (#65)

DietrichGebert:
Two-part response to #65 ("Impact on model performance?").

## Part 1 — fix the correctness gate

Two bugs in the `correct` gate were **under-reporting correctness for terse models** — the likely source of the "Ponytail degrades performance" reports.

1. **Unfenced code scored as failure.** `extractBlocks()` only matched fenced ` ``` ` blocks; bare code (common under Ponytail, frequent on `gpt-5.4-mini`) auto-failed even when correct — 41 of 74 failures in a first GPT run. Now falls back to the whole response as one block, tolerates CRLF, and detects unfenced arrow functions.
2. **Debounce tested the wrong deliverable.** Prompt said "add debounce to a search input"; the check wanted a reusable `debounce(fn, delay)` util → correct inline answers failed (`searchInput is not defined`). Reworded to the deliverable the check verifies.

Plus `correctness.test.js` (regression guard) and a README repro fix (`--env-file ../.env`).

Clean `n=20` GPT run after the fixes:

| model | baseline | ponytail | median LOC |
|---|--:|--:|--:|
| gpt-4.1-mini | 100/100 | 100/100 | 15 → 7 |
| gpt-5.4-mini | 100/100 | 98/100 | 16 → 7 |

The reported `gpt-4.1-mini` drop (10/15) does not reproduce once the gate can read the code.

## Part 2 — robustness audit

Answers the real question: does Ponytail's push for the shortest solution make weak models produce *wrong* code on edge cases?

`robustness-audit.js` — 16 self-verifying tasks (12 algorithmic edge-case traps: off-by-one, n=0, leap-century, subtractive Roman, deep nesting, … + 4 validators). Every check ships a known-good and known-lazy-wrong reference that must pass/fail before any model output is scored (`--selftest`, 16/16).

**Result (gpt-4.1-mini + gpt-5.4-mini, baseline vs ponytail):** parity on every edge-case trap, both models. The one measured soft spot is `gpt-5.4-mini` email (~4–5%, reaches for `parseaddr`). A sharpened SKILL.md validation rule had **no reliable effect** in an n=100 A/B (96% vs 95%) so it was **not shipped** — the tendency is model-level, not skill-level. Full writeup: `benchmarks/results/2026-06-16-robustness-audit.md`.

**Bottom line:** "Ponytail degrades model performance" is not supported. The LOC win (~halved) comes without a correctness tax on capable models; the lone weak-model email slip is documented honestly.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -11,9 +11,12 @@ check with `node --version` and upgrade if needed):
 
 ```bash
 cp ../.env.example ../.env      # add your ANTHROPIC_API_KEY
-npx promptfoo@latest eval -c promptfooconfig.yaml --repeat 10
+npx promptfoo@latest eval -c promptfooconfig.yaml --env-file ../.env --repeat 10
 npx promptfoo@latest view
 ```
+
+`--env-file ../.env` is required because promptfoo reads `.env` from the current
+directory (`benchmarks/`), not the repo root where the file lives.
 
 ### Local models via Ollama
 
--- a/benchmarks/claude-email.js
+++ b/benchmarks/claude-email.js
@@ -1,0 +1,40 @@
+// Email under ponytail on Claude (ponytail's primary target), baseline vs ponytail.
+const fs = require('fs'), path = require('path');
+const { checkPy, pyBlock, TASKS } = require('./robustness-audit.js');
+const skill = fs.readFileSync(path.join(__dirname, '..', 'skills', 'ponytail', 'SKILL.md'), 'utf8');
+const email = TASKS.find(t => t.name === 'email');
+const N = Number(process.env.CE_N) || 40;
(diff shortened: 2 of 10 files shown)


ryanmstokes:
hmmm the thing is, these tests are too simple. Every task is one small function you can unit-test, so it can't really go wrong. There's no abstraction to skip, no module to refactor, no error handling or edge case it might drop. The problems with "shortest answer wins" show up in multi-file, complex architectural decisions, and none of these cover that. The parseaddr slip is an example of that. At minimum you'd want to run SWE-bench with ponytail vs. baseline and see how it scores — though even that wouldn't really tell you whether it made good architectural choices, just whether the patches passed.

PR #92, opened 2026-06-16
feat: the cleanest code is no code at all

Jay-1409:
Added the below point to the ladder

0. **Can existing code be removed instead?** Deletion over addition.

Hey, 

Please have a look at this addition.

Thanks 
: )

### DIFF
--- a/.clinerules/ponytail.md
+++ b/.clinerules/ponytail.md
@@ -4,6 +4,7 @@ You are a lazy senior developer. Lazy means efficient, not careless. The best co
 
 Before writing any code, stop at the first rung that holds:
 
+0. Can existing code be removed instead?
 1. Does this need to be built at all? (YAGNI)
 2. Does the standard library already do this? Use it.
 3. Does a native platform feature cover it? Use it.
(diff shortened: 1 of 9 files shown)


DietrichGebert:
Hey Jay, thanks for taking the time on this, and for the spirit of it. "The cleanest code is no code at all" is exactly the right instinct.

I'm keeping it closed though, and I want to be upfront about why so it doesn't feel arbitrary.

Anything that changes ponytail's actual behavior (the ladder, the rules, the prompt text in the skill, hook, and mirror files) can't go in on review alone. The whole point of this repo is that the prompt earns its keep on a benchmark: less code, lower cost, same correctness. So every wording change to the ladder has to be measured before it lands, not merged because it reads well. A change that sounds better can quietly make the model behave worse, and the only way to know is to run the numbers.

On the idea itself: "deletion over addition" is already in there, down in the Rules, and rung 1 is YAGNI. A new rung 0 overlaps with both, so even on the merits I'm not sure it pays for the extra line. But that's a hunch, and hunches are exactly what the benchmark is for.

If you want to pursue it, the path is simple: run it through the benchmark, and if rung 0 beats the current ladder on the numbers, I'll take it gladly. Appreciate you digging in.

PR #101, opened 2026-06-16
docs: add translated README links

dowithless:
This PR adds links to translated versions of the README to help non-English-speaking users access translated documentation.

Supported languages:

- German
- Spanish
- French
- Japanese
- Korean
- Portuguese
- Russian
- Chinese

Translations are automatically synchronized and require no maintenance effort from project maintainers.

This change only affects README documentation and does not modify project code, dependencies, CI workflows, or repository settings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,18 @@
 
 <h1 align="center">Ponytail</h1>
 
+<!-- Keep these links, translations synced daily. -->
+<p align="center">
+<a href="https://zdoc.app/de/DietrichGebert/ponytail">Deutsch</a> |
+<a href="https://zdoc.app/es/DietrichGebert/ponytail">Español</a> |
+<a href="https://zdoc.app/fr/DietrichGebert/ponytail">français</a> |
+<a href="https://zdoc.app/ja/DietrichGebert/ponytail">日本語</a> |
+<a href="https://zdoc.app/ko/DietrichGebert/ponytail">한국어</a> |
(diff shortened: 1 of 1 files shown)


peda-cos:
Sorry, but which AI did you use for these translations? The Portuguese translation is pretty bad. I’m pretty sure even basic models like Gemma 4 can handle English to Portuguese just fine, so I’m a bit confused about which model was actually used.

dowithless:
> Sorry, but which AI did you use for these translations? The Portuguese translation is pretty bad. I’m pretty sure even basic models like Gemma 4 can handle English to Portuguese just fine, so I’m a bit confused about which model was actually used.

Thanks for the feedback! We're using deepseek-v4-flash for these translations.

peda-cos:
Interesting. It might just sound weird in Portuguese specifically. I can't really judge the other languages since I don't speak them.
I think you might need to tweak the prompt. Maybe tell it to focus on the actual meaning and translate accordingly instead of doing a literal translation. It needs a localized translation that captures the natural phrasing of the language, as if a native speaker wrote it.

DietrichGebert:
Thanks for the PR. These links point to zdoc.app, a third-party service, rather than the translations this repo already ships (README.es.md, README.ko.md, and the Chinese one). I would rather not send readers to an outside site or keep a "synced daily" note in the README, so I am going to pass on this. If you want to add a real translation, a README.<lang>.md in the repo is very welcome.

PR #107, opened 2026-06-16
Enhance ponytail mode tracker with status check

krataratha:
Added a status check feature to report the current ponytail mode.


DietrichGebert:
Thanks for this. The diff looks corrupted on my end (duplicated/unclosed blocks, the file won't parse), and the bare-/ponytail status idea is also being worked in #99. Closing this one, please reopen a clean version if you want to pursue it.

Issue #17, opened 2026-06-13
Does this support "brownfield" projects?


TomLucidor:
Assuming that there are project to be refactored without breaking, how can ponytail help in that regard? What are the risks/challenges of using PonyTail in general?

P.S. Can we use smaller models to learn from larger models' ponytail work methodology and get smarter from it?

DietrichGebert:
Thanks for opening this, good questions.

**Refactoring without breaking.** Ponytail helps two specific ways, and deliberately doesn't do a third:
- `/ponytail-review` scans a diff or a file for over-engineering and hands you a delete-list: reinvented stdlib, abstractions with one implementation, dead flexibility, config nobody reads. That's a refactoring to-do list aimed at shrinking the code, not growing it.
- Day to day it biases toward the smallest diff that works, so a refactor stays boring and reviewable instead of becoming a rewrite.
- What it does **not** give you is a safety net. "Delete this dead abstraction" is only safe if nothing depends on it, and ponytail won't run your tests or prove behavior is preserved. So keep changes small, behind a test suite, and treat its delete suggestions as candidates you verify, not commands. (Its own rules already forbid simplifying away error handling, validation, and security.)

**Risks/challenges in general.**
- It's a bias, not a guarantee. It steers the model hard toward minimal solutions; it doesn't enforce them. Review still matters.
- Under-building. YAGNI is right most of the time and wrong some of the time. When the "speculative" need was actually real, aggressive minimalism defers structure you needed. Built-in mitigations: the `lite` level (names the lazier option, you pick), the "when NOT to be lazy" carve-outs (trust-boundary validation, data-loss-preventing error handling, security, accessibility are never cut), and every shortcut leaves a `ponytail:` comment naming the upgrade path, so the ceiling is visible to the next reader.
- Trusting those carve-outs. They held across our benchmark (every arm passes the same security/concurrency probes), but on an unusual domain, sanity-check that "not lazy about" actually caught your critical path.

**P.S. small models learning the methodology.** Two layers:
- In-context, they already do. Ponytail is a prompt, not weights. Give a small model the SKILL.md as its system prompt and it applies the method immediately. The benchmark shows it: Haiku with ponytail writes ~8% of the code Haiku writes with no skill, the same proportion the bigger models get. The methodology transfers across model sizes today, no training.
[...]

PR #114, opened 2026-06-16
fix: pin all four safety carve-outs in the rule-drift canary (not just validation)

yjh8:
## What

`check-rule-copies.js` pins load-bearing rule phrases so a reword in `SKILL.md` or `AGENTS.md` can't silently drift. But of the four "never simplify away" safety carve-outs —

> input validation at trust boundaries, error handling that prevents data loss, security, accessibility

— only `input validation at trust boundaries` is in `INVARIANTS`. The other three (data-loss handling, security, accessibility) can be reworded or dropped from either file and CI stays green.

For a skill whose whole risk surface is laziness shading into negligence, those carve-outs are the guardrail. You flagged this exact thing in #17:

> Trusting those carve-outs. They held across our benchmark … but on an unusual domain, sanity-check that "not lazy about" actually caught your critical path.

This makes the drift guard cover all four, not one — so a future reword of `SKILL.md` or `AGENTS.md` that weakens a safety clause trips CI instead of slipping through.

## Change

Three phrases added to `INVARIANTS`, each verified present (verbatim, `includes()`) in **both** `skills/ponytail/SKILL.md` and `AGENTS.md` on current `main`:

- `prevents data loss`
- `security`
- `accessibility`

Note: `"error handling that prevents data loss"` as one string is **not** usable — it wraps across a line break in `SKILL.md`, so `includes()` returns false there. `prevents data loss` is the continuous substring present in both.

```
$ node scripts/check-rule-copies.js
Rule copies match AGENTS.md; 8 rule invariants present in SKILL.md and AGENTS.md.
```

No rule text changed. Smallest diff that holds.

## Optional follow-up (not in this PR)

`SKILL.md` says "security **measures**" / "accessibility **basics**" while `AGENTS.md` says "security" / "accessibility", so two of these canaries are the bare common substrings. If you'd prefer stronger, in-context canaries, harmonizing those two words across both files would let the invariant pin a distinctive phrase. Left out to keep this minimal — your call.

### DIFF
--- a/scripts/check-rule-copies.js
+++ b/scripts/check-rule-copies.js
@@ -43,7 +43,14 @@ const INVARIANTS = [
   'naive heuristic',                       // ceiling-comment rule
   'ONE runnable check',                    // test reflex
   'flimsier algorithm',                    // robust-variant rule
-  'input validation at trust boundaries',  // the "not lazy about" clause
+  // the four "not lazy about" safety carve-outs: pin each so a reword in either
+  // file can't silently drop one. Only validation was pinned before. These are the
+  // continuous substrings present in both files ("prevents data loss" because the
+  // full "error handling that prevents data loss" wraps a line in SKILL.md).
+  'input validation at trust boundaries',
+  'prevents data loss',
+  'security',
+  'accessibility',
   'Lazy code without its check is unfinished', // one-check promoted to headline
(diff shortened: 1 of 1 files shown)


salmslmalnjar-ui:
V

PR #128, opened 2026-06-17
benchmarks: add system prompt to baseline arm so it doesn't ramble (closes #126)

### LINKED ISSUE #126: Benchmark issues - baseline scores are ~7 times better
Hi, firstly, thanks for including a benchmark - most projects of this type don't have anything of the sort.

When running the benchmarks I spotted a couple of issues:

1. With the baseline test case, the model often responds in a verbose fashion with multiple options. This is because the test case lacks a system prompt that is consistent with the types of prompt used by coding harnesses
2. The debounce tests fails often for all the test targets (baseline, caveman, ponytail) due to an ambiguous test case. The response often expects the presence of a DOM (i.e. referencing `document`), which is entirely reasonable based on the test case.

Fixing (1), by adding the simple system prompt `Provide just one example for any given task, and no commentary or usage examples.`, makes quite an impact on the results:

Average LOC, running all tests (minus debounce) on haiku only:

 - Baseline (108)
 - Baseline – "one example" system prompt (16)
 - Ponytail (8.25)

As a result, the central claims of this project should be a bit more modest.

Fato07:
**What**

Add a one-line system prompt to the baseline benchmark arm so the comparison against ponytail and caveman is apples-to-apples.

**Why**

Closes #126. Without a system prompt, the baseline model often responds with multiple options and usage examples, inflating LOC ~7× for reasons unrelated to the test. The system prompt is the one Colin Eberhardt proposed in the issue and matches the style of the other arm files.

**Diff**

1 file, +6 / -2. Doesn't touch SKILL.md, AGENTS.md, the canary, or any trust-boundary code. `node scripts/check-rule-copies.js` ✅ · `npm test` ✅ (11/11).

### DIFF
--- a/benchmarks/arms/baseline.js
+++ b/benchmarks/arms/baseline.js
@@ -1,2 +1,6 @@
-// Baseline arm: no skill, just the task.
-module.exports = ({ vars }) => [{ role: 'user', content: vars.task }];
+// Baseline arm: no skill, with a one-line system prompt so the model doesn't ramble.
+const system = 'Provide just one example for any given task, and no commentary or usage examples.';
+module.exports = ({ vars }) => [
+  { role: 'system', content: system },
+  { role: 'user', content: vars.task },
+];


DietrichGebert:
Heads up @Fato07, I reverted this in #175, want to explain why since the intent was good.

A baseline arm has to be the pure model: the task and nothing else. Adding a system prompt ("provide just one example, no commentary") changes the control's behavior, which means two things break. The "baseline (no skill)" label stops being true, and the published single-shot numbers (baseline 518/693/256 LOC, the 80-94% figure) were measured against the bare baseline and never recomputed, so the code and the table no longer match.

The deeper reason: any prompt on the baseline tilts the result. "Write the minimum amount of code" would erase ponytail's gap, the opposite would inflate it. So the only honest control is no prompt at all. The rambling concern from #126 is real, but it's already answered by the agentic benchmark (that's the headline number now), so the single-shot baseline doesn't need de-rambling.

Thanks for the contribution, the instinct was right, it just can't live on the control arm.

PR #129, opened 2026-06-17
docs: correct cost claim to 42-75% from 30-rep re-verification

DietrichGebert:
## What

Re-ran the cost benchmark at **30 reps per cell** on Claude (Haiku/Sonnet/Opus) to back the headline cost number with current data. The result: ponytail is **42-75% cheaper** than no-skill, not the previously published **47-77%**. The direction holds, but both ends came in a few points lower (floor 47→42 on Opus, ceiling 77→75 on Sonnet).

## Claude, pooled 30 reps (USD, 5 tasks)

| model | baseline | ponytail | cheaper |
|---|--:|--:|--:|
| Haiku | 0.030 | 0.011 | 63% |
| Sonnet | 0.137 | 0.035 | 75% |
| Opus | 0.137 | 0.079 | 42% |

Latency 3.1-5.8x faster; ponytail correctness 100% on all three (no-skill Sonnet drops to 76% on the dict-vs-bool over-engineering bug).

## Why it stays Claude-scoped

Added OpenAI arms (gpt-4.1-mini, gpt-5.4-mini, gpt-5.5). On reasoning models the claim **reverses**: gpt-4.1-mini is 40% cheaper, but gpt-5.4-mini is 26% **more** expensive and the newest top model gpt-5.5 is **39% more expensive** and not faster. The always-on ruleset re-sent as input plus extra reasoning tokens outweighs the shorter code. So the headline stays a Claude-API claim, which the README caveat already states.

## Changes

- README headline + body, chart SVG subtitle, `benchmarks/README.md` cost table → **42-75%**.
- New dated writeup: `benchmarks/results/2026-06-17-cost-verification.md` (method, numbers, the OpenAI auto-caching gotcha that capped OpenAI cost at 10 clean reps).
- New configs: `promptfooconfig.gpt-newest.yaml`, `promptfooconfig.gemini.yaml`.

## Pending

Gemini (gemini-3.5-flash, gemini-3.1-pro-preview) hit the Google AI Studio 600/day cap mid-run; config is committed, rerun on fresh quota and append to the doc.

Reproduce:

```bash
npx promptfoo@latest eval -c benchmarks/promptfooconfig.yaml --env-file .env --repeat 10
```


Babar-maker76:
Appreciate the honesty on the single-shot caveat — "per-session cost can land either way" is exactly right, and most teams don't realize this until they're in production.

In real multi-turn sessions the ruleset re-injection compounds fast, especially if the agent loops or re-queries the same context. We've seen runs go 3-5× over the single-shot estimate in production.

This is exactly why we built runtime cost tracking into **agents.opsveritas.com** — tracks actual token spend per execution across multi-turn sessions, not just benchmark estimates. Alerts when spend hits 3× baseline before it becomes a problem.

Useful companion to ponytail's code reduction savings — you're cutting tokens at the prompt level, we catch when the savings disappear at runtime.

Ayaan-khn:
@DietrichGebert Hello Sir is there any chance we can Talk please! I am curious about this Field and have been doing my own small projects....

PR #130, opened 2026-06-17
fix(opencode): rename .opencode/command/ → .opencode/commands/ to match OpenCode's discovery convention

fcmfcm01:
## Problem

OpenCode auto-discovers custom commands from **`.opencode/commands/`** (plural), per the [official docs](https://opencode.ai/docs/commands/#markdown):

> You can also define commands using markdown files. Place them in:
> - **Global:** `~/.config/opencode/commands/`
> - **Per-project:** `.opencode/commands/`

Ponytail shipped them in **`.opencode/command/`** (singular), so the slash commands (`/ponytail`, `/ponytail-audit`, `/ponytail-review`, `/ponytail-debt`, `/ponytail-help`) were **never auto-loaded by OpenCode**. The plugin's `experimental.chat.system.transform` hook correctly injected the ruleset and persisted `/ponytail` mode switches, but typing any of those commands in the TUI produced no completion and no execution — the entire string was passed through as a user message.

The README's claim that the plugin *"adds the `/ponytail` commands"* was effectively false for everyone except someone running OpenCode from inside this repo's checkout (and even then, only by accident of the plugin-relative path).

## Fix

Rename the directory to match OpenCode's convention:

- `git mv .opencode/command .opencode/commands` (rename detected by git, file history preserved at 100% similarity)
- `tests/commands.test.js`: path assertion updated to `.opencode/commands/`
- `docs/agent-portability.md`: OpenCode row updated to `.opencode/commands/`
- `README.md`: add a paragraph in the OpenCode install section documenting how to use the commands globally across projects (symlink into `~/.config/opencode/commands/`), since the plugin path only auto-discovers commands when running OpenCode from this repo's root

## Verification

```
$ npm test
# tests 51
# pass 50
# fail 1   ← pre-existing: `csv: correct pandas one-liner passes`, unrelated (pandas not installed locally)
```

The 5 commands tests in `tests/commands.test.js` all pass after the path update. The single failure is a pre-existing CSV benchmark env issue (same `# fail 1` on `main` without my changes).

## Notes

- The directory name was singular by ponytail's own convention (`hooks/`, `skills/`, `command/` — all singular), but OpenCode's spec requires plural `commands/`. Aligning to the spec is the smaller change than asking OpenCode to accept both forms.
- Cross-project global install is now a one-liner documented in README:

```bash
mkdir -p ~/.config/opencode/commands
ln -s "$PWD"/.opencode/commands/*.md ~/.config/opencode/commands/
```

Run once from the repo root; rerun if a new command is added; restart OpenCode.


README.md
```diff
@@ -133,6 +131,15 @@ Injects the ruleset every turn at the active level; adds the `/ponytail` command
 
 The `./` path resolves against your project's `opencode.json`; to share one checkout across projects, point it at the absolute path of the `.mjs` instead (it finds its `hooks/` and `skills/` relative to its own file).
 
+The slash commands ship in `.opencode/commands/` and are auto-discovered when you run OpenCode from this repo's root. To use them in every project without copying, link them into OpenCode's global commands directory:
+
+```bash
+mkdir -p ~/.config/opencode/commands
+ln -s "$PWD"/.opencode/commands/*.md ~/.config/opencode/commands/
```
slamp, line 138: $PWD extends to where you run it, not to the root of the repository, it's easy to miss it even if it's written a few lines above. Instead, in the README we could write:
```bash
mkdir -p ~/.config/opencode/commands
ln -sf /path/to/ponytail/.opencode/commands/*.md ~/.config/opencode/commands/
```

I added the -f , because rerunning fails with "File exists" on existing symlinks

/cc @fcmfcm01

slamp: And thanks for the patch! it works in my install 

Image

slamp (changes requested):

DietrichGebert:
Thanks for digging into the OpenCode command dir. We keep it as .opencode/command/ (singular) here because that is what OpenCode loads from and the tests are green on it, so I am going to close this. Appreciate the effort.

PR #136, opened 2026-06-17
Audit-driven fixes: correctness, gate integrity, config-dir, CI/doc coverage

ferran-valvia:
## What this is

A read-only senior audit of the repo turned into a set of small, verified fixes.
Each finding was adversarially verified before landing, and every change ships
with the runnable check that fails if it regresses. No behavior is changed beyond
the bullets below; the multi-platform rule copies (the product) are untouched.

All commits keep `node scripts/check-rule-copies.js` and `npm test` green
(77 checks), and the series applies cleanly on top of `main`.

## Fixes

**Correctness / state machine**
- **mode-tracker double-output:** a prompt that both switches mode *and* contains
  "stop ponytail" / "normal mode" emitted **two JSON objects on one stdout**,
  breaking the host's `JSON.parse` (Codex/Copilot) and leaving the emitted
  `systemMessage` contradicting the final state. The deactivation branch is now
  mutually exclusive with the command branch.
- **silent level reset on a typo:** `/ponytail <unknown-arg>` fell through to
  `getDefaultMode()` and silently reset the active level in Claude Code and
  OpenCode. It is now a no-op (the active level is left alone), matching what the
  pi extension already does. Bare `/ponytail` still re-applies the default.

**Benchmark correctness gate** (`benchmarks/correctness.js`)
- **prose scored as a pass:** the unfenced-fallback block is now tagged, and the
  run-free structural checks (countdown/ratelimit) require a real code construct,
  so prose that name-drops `useState`/`FastAPI`/`429` no longer scores 1.
- **exit-0 scored as a pass:** a model demo that calls `sys.exit(0)` /
  `process.exit(0)` before the appended asserts run used to pass on exit code
  alone. Each harness already prints a `PASS` sentinel on success; the gate now
  requires it.

**Cross-platform config dir**
- The statusline scripts (`.sh` + `.ps1`) and the SessionStart setup nudge
  hardcoded `~/.claude`, but the hooks write the flag under `$CLAUDE_CONFIG_DIR`
  when set. They now mirror `getClaudeDir()`, so the badge and nudge work when the
  config dir is relocated.

**CI coverage**
- Fold the orphaned `benchmarks/correctness.test.js` (the issue #65 guard, never
  run by `npm test`/CI) into `tests/correctness.test.js`, plus regression cases
  for the two gate fixes above.
- Add a bash smoke test for the statusline badge.
- Extend the Windows hooks test to cover `copilot-hooks.json` (script existence)
  and assert `hooks.json` ↔ `copilot-hooks.json` wire the same scripts.
- Add a parity gate over the three `marketplace.json` manifests, and a
  name+description gate over the skill source frontmatter.

**Docs / metadata**
- Reconcile the host/command matrix: Antigravity reuses `gemini-extension.json`
  (same as Gemini CLI) and surfaces commands as chat-skills, so it is documented
  as command-capable; Copilot CLI and OpenClaw are marked command-capable and
  "Copilot (editor)" as the instruction-only one; the bare `/ponytail` row is
  corrected (it re-applies the default, it does not report the level).
- Document `.agents/plugins/marketplace.json` and add the missing Aider row.
- Drop the unused `license: MIT` frontmatter key from `skills/ponytail`.

## Deliberately not changed
- `mode-tracker` still accepts a `$ponytail` prefix. It is undocumented and no
  known host emits it, but narrowing an input parser on an unverified assumption
  risks breaking a niche host for a cosmetic win, so it is left as-is.

## Limits
- Antigravity's exact command surface and the precise CLI verb that resolves
  `.agents/plugins/marketplace.json` were reasoned from the shared manifest and
  docs, not confirmed against a live host.
- CI is ubuntu-only, so the PowerShell statusline path is static-checked, not
  executed; the bash statusline smoke test does run.


DietrichGebert:
Thanks for the audit work here, there's real signal in it. The tricky part is that since this was opened most of these fixes have landed on main separately: the statusline config-dir fix (#154), the activate.js nudge path, the mode-tracker double-output guard, and the opencode typo no-op are all in already, and the branch is conflicting now. The one piece I don't see on main is the benchmark gate-integrity change in benchmarks/correctness.js (the PASS sentinel check plus the prose rejection). Could you reopen a small focused PR with just that part rebased on main? That one I'd love to have. Closing this as mostly superseded. Thanks!

Issue #34, opened 2026-06-14
hooks/ponytail-activate.js does not honor CLAUDE_CONFIG_DIR


AbhishekSharma20:
const claudeDir = path.join(os.homedir(), '.claude');

This line in hooks/ponytail-activate.js assumes claudeDir as homedir irrespective of the env var CLAUDE_CONFIG_DIR which the claude uses to override claude dir if present else fal black to ~/.claude

Lewiscowles1986:
So switching to the following?

`process.env.CLAUDE_CONFIG_DIR?.strip() || path.join(os.homedir(), '.claude');`

Lewiscowles1986:
Yo @DietrichGebert the above good enough?
_cascade to use the environment var, and only default to `$HOME/.claude` if it's not set?_

DietrichGebert:
Fixed and shipped in v4.3.0. The hooks now resolve the dir through a shared `getClaudeDir()` that returns `CLAUDE_CONFIG_DIR` when set and falls back to `~/.claude` (#37), with a regression test.

@Lewiscowles1986 thanks for the nudge, I went with the plain `CLAUDE_CONFIG_DIR || ~/.claude` instead of trimming, since a config path with surrounding whitespace would be unusual, but that's easy to add if it ever bites. Thanks @AbhishekSharma20 for the clear report.

PR #154, opened 2026-06-18
fix: statusline honors CLAUDE_CONFIG_DIR (completes #34)

### LINKED ISSUE #153: Statusline badge ignores CLAUDE_CONFIG_DIR — shows nothing or a stale mode for non-default config homes
### Version

v4.7.0 (tag v4.7.0)

### Problem

The mode flag and the statusline disagree on **where** the flag lives.

Issue #34 made the hooks honor `CLAUDE_CONFIG_DIR` when **writing** the flag. `getClaudeDir()` resolves it like Claude Code does:

```js
// hooks/ponytail-config.js
function getClaudeDir() {
  // CLAUDE_CONFIG_DIR overrides ~/.claude, matching Claude Code.
  return process.env.CLAUDE_CONFIG_DIR || path.join(os.homedir(), '.claude');
}
```

and `tests/hooks.test.js` explicitly asserts the flag is written to `$CLAUDE_CONFIG_DIR/.ponytail-active` (and *not* `~/.claude`) when that env var is set.

But **both statusline scripts still read the old hardcoded path** and never consult `CLAUDE_CONFIG_DIR`:

```bash
# hooks/ponytail-statusline.sh
flag="$HOME/.claude/.ponytail-active"
```

```powershell
# hooks/ponytail-statusline.ps1
$Flag = Join-Path $HOME ".claude/.ponytail-active"
```

### Impact

For any user who sets `CLAUDE_CONFIG_DIR` (multi-account, sandboxed, or non-default config home — a first-class Claude Code setting):

- The activation hook writes the mode to `$CLAUDE_CONFIG_DIR/.ponytail-active`.
- The statusline reads `$HOME/.claude/.ponytail-active` — a different file.

Result: the badge shows **nothing**, or — if the user previously ran without `CLAUDE_CONFIG_DIR` — a **stale mode** from the old `~/.claude` flag that will never update again, silently misreporting the active level.

The write half of #34 was fixed and tested; the read half (the statusline) was left pointing at the old location. The header comment in `ponytail-activate.js` ("Writes flag file at ~/.claude/.ponytail-active") is now inaccurate too.

### Fix

Resolve the flag in the statusline scripts the same way `getClaudeDir()` does — prefer `CLAUDE_CONFIG_DIR`, fall back to `~/.claude`:

```bash
flag="${CLAUDE_CONFIG_DIR:-$HOME/.claude}/.ponytail-active"
```

```powershell
$ClaudeDir = if ($env:CLAUDE_CONFIG_DIR) { $env:CLAUDE_CONFIG_DIR } else { Join-Path $HOME ".claude" }
$Flag = Join-Path $ClaudeDir ".ponytail-active"
```

The fallback branch is identical to today's behavior, so users without the env var set are unaffected.

### Severity

MEDIUM — statusline-only (core laziness behavior is unaffected), but it breaks the badge for every `CLAUDE_CONFIG_DIR` user and the stale-mode case actively misinforms. It completes a fix (#34) that was only half-applied.

Lakshya77089:
Closes #153.

## Summary

Issue #34 made the hooks honor `CLAUDE_CONFIG_DIR` when **writing** the mode flag, and [tests/hooks.test.js](tests/hooks.test.js#L81-L100) enforces it lands in `$CLAUDE_CONFIG_DIR/.ponytail-active`. But both statusline scripts still **read** a hardcoded `$HOME/.claude/.ponytail-active`:

- [ponytail-statusline.sh:2](hooks/ponytail-statusline.sh#L2)
- [ponytail-statusline.ps1:1](hooks/ponytail-statusline.ps1#L1)

So for any user with `CLAUDE_CONFIG_DIR` set, the write side and read side point at different files: the badge shows nothing, or a **stale mode** from a pre-migration `~/.claude` flag that never updates. The write half of #34 was fixed; the read half was missed.

## Fix

Resolve the flag in both scripts exactly as `getClaudeDir()` does — prefer `CLAUDE_CONFIG_DIR`, fall back to `~/.claude`:

```bash
flag="${CLAUDE_CONFIG_DIR:-$HOME/.claude}/.ponytail-active"
```
```powershell
$ClaudeDir = if ($env:CLAUDE_CONFIG_DIR) { $env:CLAUDE_CONFIG_DIR } else { Join-Path $HOME ".claude" }
$Flag = Join-Path $ClaudeDir ".ponytail-active"
```

The fallback branch is byte-for-byte the previous behavior, so users without the env var are unaffected. Also corrects the now-inaccurate path comment in the activation hook header.

## Verification

- `bash -n` and PowerShell parser both pass.
- Functionally checked both scripts: with `CLAUDE_CONFIG_DIR` set they read the custom dir (`[PONYTAIL:ULTRA]`); unset, they read `~/.claude` (`[PONYTAIL:LITE]`).
- Full suite green: `node --test tests/*.test.js` → 51/51, and `node scripts/check-rule-copies.js` passes.

### DIFF
--- a/hooks/ponytail-activate.js
+++ b/hooks/ponytail-activate.js
@@ -2,7 +2,7 @@
 // ponytail — Claude Code SessionStart activation hook
 //
 // Runs on every session start:
-//   1. Writes flag file at ~/.claude/.ponytail-active (statusline reads this)
+//   1. Writes flag file at $CLAUDE_CONFIG_DIR/.ponytail-active (defaults to ~/.claude; statusline reads this)
 //   2. Emits ponytail ruleset as hidden SessionStart context
 //   3. Detects missing statusline config and emits setup nudge
 
--- a/hooks/ponytail-statusline.ps1
+++ b/hooks/ponytail-statusline.ps1
@@ -1,4 +1,6 @@
-$Flag = Join-Path $HOME ".claude/.ponytail-active"
+# CLAUDE_CONFIG_DIR overrides ~/.claude, matching where the hooks write the flag (issue #34)
+$ClaudeDir = if ($env:CLAUDE_CONFIG_DIR) { $env:CLAUDE_CONFIG_DIR } else { Join-Path $HOME ".claude" }
+$Flag = Join-Path $ClaudeDir ".ponytail-active"
 if (-not (Test-Path $Flag)) {
     exit 0
 }
--- a/hooks/ponytail-statusline.sh
+++ b/hooks/ponytail-statusline.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
-flag="$HOME/.claude/.ponytail-active"
+# CLAUDE_CONFIG_DIR overrides ~/.claude, matching where the hooks write the flag (issue #34)
+flag="${CLAUDE_CONFIG_DIR:-$HOME/.claude}/.ponytail-active"
 [ -f "$flag" ] || exit 0
 
 mode=$(head -n1 "$flag" | tr -d '[:space:]')


huzaif-fahad (commented):
LGTM!

PR #170, opened 2026-06-18
fix(benchmarks): portable plugin-dir resolution for agentic arms

### LINKED ISSUE #169: Agentic benchmark: hardcoded personal plugin-cache paths break reproducibility of ponytail/caveman arms
### Summary

`benchmarks/agentic/run.py` hardcodes the maintainer's personal Windows plugin-cache paths, so the `ponytail` and `caveman` arms of the new agentic benchmark only run on one machine. This undercuts the "fully reproducible" claim that the rebuilt benchmark (see #126, and the Scott Logic critique it answers) was specifically meant to establish.

### Where

https://github.com/DietrichGebert/ponytail/blob/main/benchmarks/agentic/run.py (lines ~47-50)

```python
PLUGIN_DIRS = {
    "ponytail": r"C:\Users\Dietr\.claude\plugins\cache\ponytail\ponytail\4.2.0",
    "caveman":  r"C:\Users\Dietr\.claude\plugins\cache\caveman\caveman\63e797cd753b",
}
```

`PLUGIN_DIRS[arm]` is passed to `--plugin-dir` (line ~221). These absolute paths point at `C:\Users\Dietr\...` and pin a version (`4.2.0`) and content hash (`63e797cd753b`) that exist only in that user's cache.

### Impact

- Anyone else running `benchmarks/agentic/run.py` for the `ponytail` or `caveman` arms gets a non-existent `--plugin-dir`; the smoke test (which verifies activation by output style) fails, so the skill is never actually activated.
- Only `baseline`, `yagni`, and `yagni-oneliner` (raw `--append` prompts) are reproducible off the maintainer's box.
- The headline result ("ponytail writes ~54% less code", the safety-tier comparison) cannot be independently reproduced by a third party — which is the exact bar #126 raised.

### Proposed fix (portable, minimal)

Resolve the plugin dir at use-site instead of hardcoding a home path:

1. Env override first: `PONYTAIL_PLUGIN_DIR` / `CAVEMAN_PLUGIN_DIR`.
2. Else discover under `Path.home() / ".claude/plugins/cache/<plugin>/<plugin>/"`, picking the latest version dir via glob (no pinned version/hash).
3. Else `sys.exit` with a clear message naming the env var to set.

Resolved only for the arm actually being run, so a missing `caveman` install doesn't block a `ponytail`-only run. Happy to open a PR with this + a small selftest.

### Context

This is a follow-up to #126 (the benchmark-fairness thread with @ColinEberhardt) and the Scott Logic post "Ponytail? YAGNI!" (https://blog.scottlogic.com/2026/06/16/ponytail-yagni-and-the-problem-with-prompt-benchmarks.html). The methodology rebuild already landed; this is the remaining reproducibility gap in that rebuild.

ousamabenyounes:
Fixes #169.

## Problem

`benchmarks/agentic/run.py` hardcoded the maintainer's personal Windows plugin-cache paths:

```python
PLUGIN_DIRS = {
    "ponytail": r"C:\Users\Dietr\.claude\plugins\cache\ponytail\ponytail\4.2.0",
    "caveman":  r"C:\Users\Dietr\.claude\plugins\cache\caveman\caveman\63e797cd753b",
}
```

Passed to `--plugin-dir`, these only exist on one machine. Everyone else's `ponytail`/`caveman` arms get a non-existent dir, the activation smoke test fails, and the skill never loads — so the headline 54% result and the safety-tier comparison can't be independently reproduced. That is the exact "fully reproducible" bar the rebuilt benchmark (#126, answering the Scott Logic critique) set out to meet.

## Fix

Resolve each plugin dir at use-site, portably:

1. Env override: `PONYTAIL_PLUGIN_DIR` / `CAVEMAN_PLUGIN_DIR`.
2. Else latest version dir under `~/.claude/plugins/cache/<plugin>/<plugin>/` (glob — no pinned version or content hash).
3. Else `sys.exit` naming the env var to set.

Resolved only for the arm actually run, so a missing `caveman` install can't block a `ponytail`-only run. No hardcoded paths, explicit filesystem error handling.

## Test verification (RED → GREEN)

Extended the existing `--selftest` with two checks: env-override honored, and missing-install fails loudly (`sys.exit`).

**RED** — resolver reverted to the old hardcoded behavior, test kept:
```
XX  plugin_dir   env  override honored
XX  plugin_dir   miss clear error (sys.exit)
selftest: 2 BROKEN
```

**GREEN** — with the fix:
```
ok  plugin_dir   env  override honored
ok  plugin_dir   miss clear error (sys.exit)

selftest: all instruments valid
```

### DIFF
--- a/benchmarks/agentic/run.py
+++ b/benchmarks/agentic/run.py
@@ -22,7 +22,7 @@ over-engineering score is a later pass.
 ponytail: the claude CLI is the harness (already installed, we run inside it). No SDK
 dependency. The CLI's JSON output already carries cost/tokens/duration/permission_denials.
 """
-import argparse, concurrent.futures, datetime, json, re, shutil, statistics, subprocess, sys, tempfile
+import argparse, concurrent.futures, datetime, json, os, re, shutil, statistics, subprocess, sys, tempfile
 from collections import defaultdict
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


DietrichGebert:
This is exactly what I wanted. Ran the selftest, env override and the loud exit both pass, existing stuff still green. Clean.

One tiny thing for later: `sorted(...)[-1]` for "latest version" is lexical not semver, so if the cache ever has two versions, `4.9.0` beats `4.10.0` and you grab the wrong one. Cache usually has one dir so it basically never happens, not a blocker. Either `max()` with a version key or a `ponytail:` comment naming the limit, whenever you feel like it.

Merging as-is. Thanks for this.

Issue #168, opened 2026-06-18
issue when using it on copilot windows


darklight9811:
<img width="443" height="199" alt="Image" src="https://github.com/user-attachments/assets/02f2def9-d896-43ed-a7af-43a41e85d63a" />

the error translates to `the token is not a valid instructions separator in this version`

DietrichGebert:
Thanks for the report and the screenshot.

That error ("the token is not a valid instructions separator in this version") looks like the Copilot CLI build rejecting the format the SessionStart hook uses to inject context, not ponytail's content itself.

Two things would help me pin it:

1. Which Copilot CLI version (`copilot --version`)?
2. Does the instruction-only fallback work for you? Drop `AGENTS.md` or `.github/copilot-instructions.md` in the project, or copy the rules into `~/.copilot/copilot-instructions.md`. That path skips the hook entirely, so if it works it confirms the hook output is the culprit.

Leaving open while I dig in.

maxfelker:
<img src="https://raw.githubusercontent.com/maxfelker/ponytail/main/assets/logo.png" width="64" height="64" />

**tl;dr (ponytail, on behalf of @maxfelker):** the hook is the bug, not the rules. skip it. `copilot-instructions.md` is the fix right now.

---

**ponytail audit of this issue:**

`yagni:` the SessionStart hook is convenience scaffolding, not load-bearing. It injects context ponytail's rules don't need at runtime — they're static text. This Copilot version is rejecting the separator format the hook emits. The rules themselves are fine.

`native:` `.github/copilot-instructions.md` is the platform-native path. It bypasses the hook entirely, requires zero code change, and works on every version that supports instructions files.

**shortest fix:**
```
cp ~/.claude/plugins/.../ponytail/rules.md .github/copilot-instructions.md
```
(or copy the relevant rules block in manually)

If the hook separator format needs patching for your Copilot version, `copilot --version` will tell us which token format it expects — but try the instructions file first. The lazy path usually works.

maxfelker:
@darklight9811 thanks for the issue report. Would love to hear what version you're on and if the @DietrichGebert recommendations work.

@DietrichGebert had ponytail run an audit on the and it confirmed your recommendation. I'm going debug locally in the mean time to see if I can reproduce

maxfelker:
[...]

PR #181, opened 2026-06-19
Fix for #168: Don't write output on SessionStart for Copilot

maxfelker:
**tl;dr:** `!isCodex` was too broad — Copilot sessions leaked Claude-only content into `additionalContext`, breaking token validation on Windows.

---

## What broke

`ponytail-activate.js` had two guards that used `isCodex` to skip Claude-specific behaviour. Both conditions forgot about Copilot — so Copilot sessions fell through into Claude-only paths.

**Bug 1 — statusline nudge injected into Copilot context**

```js
// before
if (!isCodex) try {
```

The statusline nudge appends a raw JSON-like fragment to the hook output:

```
"statusLine": { "type": "command", "command": "powershell -ExecutionPolicy Bypass -File \"...\"" }
```

Copilot's parser received this inside `additionalContext`, tried to interpret it as a structured instruction token, and rejected it: *"the token is not a valid instructions separator in this version."*

The statusline is a Claude Code setting (`~/.claude/settings.json`). Copilot has no equivalent and should never see this nudge.

**Bug 2 — off-mode emitted `'OK'` to Copilot**

```js
// before
writeHookOutput('SessionStart', 'off', isCodex ? '' : 'OK');
```

Codex got `''` → empty JSON output `{}`. Copilot got `'OK'` → `{ "additionalContext": "OK" }`. In off mode nothing should be injected. Same class of bug, same fix.

## What changed

`hooks/ponytail-activate.js` only — two lines:

```js
// import isCopilot alongside isCodex
const { clearMode, isCodex, isCopilot, setMode, writeHookOutput } = require('./ponytail-runtime');

// Bug 1: scope statusline nudge to Claude only
if (!isCodex && !isCopilot) try {

// Bug 2: off mode emits nothing for both non-Claude hosts
const hookOutput = (isCodex || isCopilot) ? '' : 'OK';
writeHookOutput('SessionStart', 'off', hookOutput);
```

`isCopilot` was already exported from `ponytail-runtime.js` — it just wasn't imported here.

## Test

`node tests/hooks.test.js` — hook compatibility checks passed.

---

*on behalf of @maxfelker — ponytail found it, two lines fixed it.*

### DIFF
--- a/hooks/ponytail-activate.js
+++ b/hooks/ponytail-activate.js
@@ -13,6 +13,7 @@ const { getPonytailInstructions } = require('./ponytail-instructions');
 const {
   clearMode,
   isCodex,
+  isCopilot,
   setMode,
   writeHookOutput,
 } = require('./ponytail-runtime');
@@ -40,7 +41,7 @@ try {
 let output = getPonytailInstructions(mode);
 
 // 3. Detect missing statusline config — nudge Claude to help set it up
-if (!isCodex) try {
+if (!isCodex && !isCopilot) try {
   let hasStatusline = false;
   if (fs.existsSync(settingsPath)) {
     // Strip UTF-8 BOM some editors prepend on Windows (breaks JSON.parse)
--- a/hooks/ponytail-activate.js
+++ b/hooks/ponytail-activate.js
@@ -26,7 +26,8 @@ const mode = getDefaultMode();
 // "off" mode — skip activation entirely, don't write flag or emit rules
 if (mode === 'off') {
   clearMode();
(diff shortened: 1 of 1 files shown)


rajpratham1 (approved):
Nice improvement! Adding `isCopilot` alongside the existing `isCodex` checks keeps the behavior consistent across supported environments. I also like the introduction of the `hookOutput` variable—it makes the intent a bit clearer and avoids repeating the conditional. Overall, the changes look clean and make the logic easier to follow. LGTM!

PR #187, opened 2026-06-19
fix: skip the Claude Code statusline nudge for Copilot

### LINKED ISSUE #186: Copilot CLI users get a bogus "set up ~/.claude statusline" nudge every session
The SessionStart activation hook injects a statusline-setup nudge into Copilot CLI sessions, pointing at a Claude Code-only setting that does nothing for Copilot.

`ponytail-activate.js` gates the nudge with `if (!isCodex)`, which lets Copilot through:

```js
// hooks/ponytail-activate.js
if (!isCodex) try {
  ...
  if (!hasStatusline) {
    output += "STATUSLINE SETUP NEEDED: ... add this to ~/.claude/settings.json: " +
      statusLineSnippet + " Proactively offer to set this up for the user on first interaction.";
  }
}
```

The statusline badge is a Claude Code feature (configured in `~/.claude/settings.json`). Copilot CLI has no such statusline, so for every Copilot user without that file the hook tells the agent to *proactively offer* to edit a Claude Code settings file that is irrelevant to them.

Reproduced by running the hook under a Copilot env:

```
COPILOT_PLUGIN_DATA=/tmp/x PONYTAIL_DEFAULT_MODE=full node hooks/ponytail-activate.js
```

The emitted `additionalContext` contains `STATUSLINE SETUP NEEDED`, the `~/.claude/settings.json` snippet, and the "Proactively offer to set this up" instruction.

Codex is already excluded. Copilot should be too — the nudge only makes sense for Claude Code. Fix is to gate it with `if (!isCodex && !isCopilot)`. PR incoming.

Lakshya77089:
Fixes #186.

The SessionStart activation hook gated the statusline setup nudge with `if (!isCodex)`, which let Copilot CLI through. The statusline badge is a Claude Code feature (configured in `~/.claude/settings.json`); Copilot has no equivalent, so every Copilot user without that file got injected guidance telling the agent to *proactively offer* to set up an irrelevant `~/.claude/settings.json` entry. Repro in the issue.

Gate the nudge with `if (!isCodex && !isCopilot)` so it only fires for Claude Code. The flag-write and ruleset injection are untouched — Copilot still gets the ponytail ruleset, just not the bogus statusline nudge.

- `hooks/ponytail-activate.js`: import `isCopilot`, exclude it from the statusline block
- tests: assert Copilot SessionStart output no longer contains the nudge, and that Claude Code with no statusline configured still gets it

`node --test tests/*.test.js` passes (51).

### DIFF
--- a/hooks/ponytail-activate.js
+++ b/hooks/ponytail-activate.js
@@ -13,6 +13,7 @@ const { getPonytailInstructions } = require('./ponytail-instructions');
 const {
   clearMode,
   isCodex,
+  isCopilot,
   setMode,
   writeHookOutput,
 } = require('./ponytail-runtime');
@@ -39,8 +40,10 @@ try {
 // 2. Emit the ponytail ruleset, filtered to the active intensity level.
 let output = getPonytailInstructions(mode);
 
-// 3. Detect missing statusline config — nudge Claude to help set it up
-if (!isCodex) try {
+// 3. The statusline badge is a Claude Code feature configured in ~/.claude.
+// Skip the setup nudge for Codex and Copilot — neither reads that statusline,
+// so nudging them points at an irrelevant ~/.claude/settings.json.
+if (!isCodex && !isCopilot) try {
   let hasStatusline = false;
   if (fs.existsSync(settingsPath)) {
     const settings = JSON.parse(fs.readFileSync(settingsPath, 'utf8'));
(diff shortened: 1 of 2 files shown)


Lakshya77089:
Closing as superseded. Main already gates the SessionStart statusline nudge for Copilot via #181 (`if (!isCodex && !isCopilot)` in `ponytail-activate.js`), which also covers the off-mode output path. I branched this off a stale `origin/main` and missed that the fix had already landed, so this just duplicates #181 and now conflicts. No action needed.

PR #191, opened 2026-06-19
Add Modern Web Guidance as an optional examples/ note for web tasks

liady:
Surfaces Modern Web Guidance as the rung-3 lookup for web work, kept entirely out of the always-on core skill per maintainer feedback.

- examples/web-platform-lookup.md: a <dialog closedby> vs Radix before/after in the date-picker.md style, carrying the lookup mechanic, the feature-type list, and the "optional, skip if the CLI isn't installed" framing.
- README.md: one "Pairs well with" line by the existing Caveman cross-reference, pointing at the example.

SKILL.md is left byte-identical to upstream, so the always-on ruleset stays lean and self-contained and there is no skill mirror to regenerate. check-rule-copies.js passes; the only test failure is the environmental pandas CSV check documented in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,8 @@ With ponytail:
 
 More survivors in [examples/](examples/).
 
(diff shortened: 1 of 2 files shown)


DietrichGebert:
Really nicely written, and it matches the ponytail spirit. I would rather not put a pairing with an outside tool at the top of the README or lean the ladder on a CLI most people will not have installed, so I am going to pass on this one. Thanks for taking the time.

PR #207, opened 2026-06-20
feat: add native Cursor plugin adapter

harshav167:
## Summary
- Add a Cursor plugin manifest that exposes Ponytail skills, existing Cursor rules, and Cursor-native hooks.
- Add Cursor hook adapters for session activation, mode tracking, dependency-install approval, compaction reminders, and session cleanup.
- Preserve existing Claude/Codex/Copilot hook behavior while adding Cursor output/state handling.

## Test plan
- node tests/hooks.test.js
- node tests/opencode-plugin.test.js
- node scripts/check-rule-copies.js
- node tests/hooks-windows.test.js
- node tests/gemini-extension.test.js
- node tests/behavior.test.js
- node tests/correctness.test.js
- Node smoke checks for Cursor hook JSON outputs


DietrichGebert:
Thanks for the Cursor adapter. This does more than port the ruleset though, it adds a shell hook that gates installs, which is new behavior and a bigger product call than I want to take on right now. Going to pass, but I appreciate the work.

PR #210, opened 2026-06-20
feat: add Aider support via .aider.conf.yml

uddin-rajaul:
Aider is listed in the README as an instruction-tier agent but had no dedicated config file. Adds .aider.conf.yml that points aider at the existing AGENTS.md ruleset, install section in README, portability table entry, and .gitignore for aider artifacts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,5 +22,8 @@ ponytail-*.gif
 # Claude Code local settings (machine-specific permission grants)
 .claude/settings.local.json
 
+# Aider temp files / chat history
+.aider*
+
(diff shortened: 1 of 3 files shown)


lukeaus:
@uddin-rajaul can you unlink #210 from this PR, this only adds Aider support, not Factory Droid. Thanks

uddin-rajaul:
@lukeaus done

DietrichGebert:
Thanks for adding Aider. It is already in the rules list on main, and this PR points to a .aider.conf.yml that it does not actually add, so it would not work as written. Closing this one, but a small follow up that ships the config file would be welcome.

PR #214, opened 2026-06-20
fix: Windows UserPromptSubmit hook hangs indefinitely on stdin

sbqb2183:
## Problem

On Windows, the `UserPromptSubmit` hook causes Claude Code to hang after every prompt — the user sends a message and gets no response indefinitely.

**Root cause:** The `commandWindows` in `claude-codex-hooks.json` wraps the node call in a PowerShell script-block:

```powershell
if (Get-Command node -ErrorAction SilentlyContinue) { node ...ponytail-mode-tracker.js }
```

PowerShell **does not forward stdin to child processes inside `if { }` script-blocks**. So `ponytail-mode-tracker.js` starts, opens stdin, and waits for the `end` event that never comes — blocking Claude Code from processing the prompt until the 5-second hook timeout kills the process.

## Fix

Two changes:

**1. `hooks/claude-codex-hooks.json`** — call `node` directly in `commandWindows` (no `if`-block wrapper). This lets the shell pipe stdin through correctly. Node is already a hard requirement so the availability check is unnecessary.

**2. `hooks/ponytail-mode-tracker.js`** — add a 500 ms hard-exit timeout as a belt-and-suspenders safety net. A `done` guard prevents double-processing if stdin closes before the timer fires. This keeps the hook well within the 5-second timeout on any platform.

## Verified

Diagnosed and fixed on Windows 10 / Claude Code desktop app v2.1.183. Both hooks now exit in ~80 ms.

## Test plan

- [ ] Send a prompt in Claude Code on Windows — response arrives immediately
- [ ] `/ponytail lite` still switches mode correctly
- [ ] `stop ponytail` still deactivates
- [ ] Mac/Linux behavior unchanged (Unix `command` path untouched)

### DIFF
--- a/hooks/claude-codex-hooks.json
+++ b/hooks/claude-codex-hooks.json
@@ -7,7 +7,7 @@
           {
             "type": "command",
             "command": "command -v node >/dev/null 2>&1 && node \"${CLAUDE_PLUGIN_ROOT}/hooks/ponytail-activate.js\" || exit 0",
-            "commandWindows": "if (Get-Command node -ErrorAction SilentlyContinue) { node \"$env:CLAUDE_PLUGIN_ROOT\\hooks\\ponytail-activate.js\" }",
+            "commandWindows": "node \"$env:CLAUDE_PLUGIN_ROOT\\hooks\\ponytail-activate.js\"",
             "timeout": 5,
             "statusMessage": "Loading ponytail mode..."
           }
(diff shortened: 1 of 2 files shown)


DietrichGebert:
Thanks for chasing this hang. Main already got the same stdin fix a while back, an error handler plus a short timeout fallback, so this one is covered now. Nice catch though.

sbqb2183:
Thx for the feedback you doing an amazing job 👏

PR #220, opened 2026-06-20
doc: add missing codex plugin installation command

kumanday:
After adding the marketplace, you need to issue another command to actually install the plugin.


DietrichGebert:
The codex plugin add line is now in the README via #545, so closing. The Spanish README still misses it; if you want, that one-liner would be a welcome follow-up. Thanks!

PR #230, opened 2026-06-21
Add Kimi Code CLI adapter

stoyanov-kaloyan:
This PR adds a Kimi Code CLI adapter for Ponytail.

What changed:
- Added `.agents/skills/ponytail*/SKILL.md` skill package generated from the canonical `skills/` directory.
- Added `scripts/build-kimi-skills.js` to regenerate the package.
- Added `tests/kimi-skills.test.js` to guard against drift.
- Updated `scripts/check-rule-copies.js` to verify rule invariants in the Kimi skill copy.
- Updated `docs/agent-portability.md` and `README.md` with Kimi install instructions.

Kimi reads `AGENTS.md` at the project root for always-on rules and discovers `.agents/skills/` for skill invocation, so the adapter follows the same thin-adapter pattern already used for OpenClaw and the instruction-only hosts.

Tested: `node scripts/check-rule-copies.js` passes and the new Kimi skill tests pass. The single unrelated failure in `tests/correctness.test.js` (missing `pandas`) is pre-existing.

### DIFF
--- a/.agents/skills/ponytail-audit/SKILL.md
+++ b/.agents/skills/ponytail-audit/SKILL.md
@@ -1,0 +1,37 @@
+---
+name: ponytail-audit
+description: "Audit the whole repo for over-engineering. A ranked list of what to delete, simplify, or replace with stdlib or native features."
+homepage: https://github.com/DietrichGebert/ponytail
+license: MIT
+---
+
+ponytail-review, repo-wide. Scan the whole tree instead of a diff. Rank
+findings biggest cut first.
+
+## Tags
+
+Same as ponytail-review:
+
+- `delete:` dead code, unused flexibility, speculative feature. Replacement: nothing.
+- `stdlib:` hand-rolled thing the standard library ships. Name the function.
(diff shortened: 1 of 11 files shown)


DietrichGebert:
Thanks for this, and SKILL.md directories are the right mechanism for Kimi. The catch is it copies the six skills into `.agents/skills/` with a generator, but the repo already has them under `skills/`, and Kimi can load those directly, so the duplicate tree and generator are more than Kimi needs (and the branch conflicts now too). I am going to document Kimi as a one-line note instead and close this. Really appreciate the careful test-guarded copy work though.

For reference, using ponytail with Kimi Code CLI with no new files:

- **Rules:** Kimi reads `AGENTS.md` from the project root, so this repo's `AGENTS.md` is the always-on ruleset.
- **Skills:** `kimi --skills-dir skills/` loads the six bundled skills (`/ponytail`, `/ponytail-review`, and the rest), or add `skills/` to `extra_skill_dirs` to keep it on by default.

Thanks again.

PR #242, opened 2026-06-22
[codex] harden benchmark subprocess execution

killfyvibecoding:
## Summary

- run benchmark subprocess checks through `execFileSync` with argv instead of shell-built command strings
- make the CSV correctness test pass without requiring optional local pandas
- keep Copilot lifecycle hooks quiet when `node` is missing, matching the Claude/Codex hook behavior

## Why

The local correctness suite failed when pandas was not installed, even though pandas is documented as an optional local dependency for CSV checks. While investigating, the benchmark scripts also had avoidable shell command construction around generated temp files. This patch hardens those execution paths and adds regression coverage.

## Validation

- `rtk npm test`
- `rtk npm test --prefix ponytail-mcp`
- `rtk node scripts/check-rule-copies.js`
- `rtk node benchmarks/robustness-audit.js --selftest`
- `rtk git diff --check`

### DIFF
--- a/benchmarks/correctness.js
+++ b/benchmarks/correctness.js
@@ -6,7 +6,7 @@
 // Unlike loc.js (measurement-only), this one is a gate — a wrong answer is a
 // wrong answer regardless of how few lines produced it.
 
-const { execSync } = require('child_process');
+const { execFileSync } = require('child_process');
 const fs = require('fs');
 const os = require('os');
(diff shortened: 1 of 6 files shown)


DietrichGebert:
Closing this draft as stale. Feel free to reopen as a regular PR if you want to continue the benchmark hardening.

PR #251, opened 2026-06-22
fix: point statusline setup nudge at CLAUDE_CONFIG_DIR/settings.json

### LINKED ISSUE #250: Statusline setup nudge hardcodes ~/.claude/settings.json, ignoring CLAUDE_CONFIG_DIR

### Where
`hooks/ponytail-activate.js`:

### Impact

### Fix

PR incoming.

halindrome:
## Problem

Issue #34 made `ponytail-activate.js` honor `CLAUDE_CONFIG_DIR` for the flag file and for *detecting* a missing statusline — it reads `settings.json` from `getClaudeDir()`. But the user-facing **statusline setup nudge** text still hardcoded `~/.claude/settings.json` in both branches (shell-safe and manual).

So with `CLAUDE_CONFIG_DIR` set, ponytail correctly detects the missing statusline in the custom dir, then tells the user (and the agent) to write the `statusLine` config into `~/.claude/settings.json` — the wrong file. The statusline never activates.

Fixes #250.

## Change

- `hooks/ponytail-activate.js`: interpolate the already-computed `settingsPath` (`path.join(getClaudeDir(), 'settings.json')`) into both nudge strings instead of the hardcoded literal. No new variables, no behavior change when `CLAUDE_CONFIG_DIR` is unset (path resolves to `<home>/.claude/settings.json`).
- `tests/hooks.test.js`: extend the existing `CLAUDE_CONFIG_DIR` case to assert the nudge references `CLAUDE_CONFIG_DIR/settings.json` and does **not** contain `~/.claude/settings.json`.

## Testing

`npm test` — all 12 checks pass, including the new assertions.

### DIFF
--- a/hooks/ponytail-activate.js
+++ b/hooks/ponytail-activate.js
@@ -66,7 +66,7 @@ if (!isCodex && !isCopilot) try {
       output += "\n\n" +
         "STATUSLINE SETUP NEEDED: The ponytail plugin includes a statusline badge showing active mode " +
         "(e.g. [PONYTAIL], [PONYTAIL:ULTRA]). It is not configured yet. " +
-        "To enable, add this to ~/.claude/settings.json: " +
+        "To enable, add this to " + settingsPath + ": " +
         statusLineSnippet + " " +
         "Proactively offer to set this up for the user on first interaction.";
     } else {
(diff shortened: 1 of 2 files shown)


DietrichGebert:
Thanks for this. It is identical to #256, which you also opened (same diff, same test). Closing as a duplicate and tracking the change in #256.

PR #259, opened 2026-06-23
Split 2 plugin install commands in README.md

grepsedawk:
Allows users to use the copy button to the left on the 2 individual cmds If you copy them together, the command fails


DietrichGebert:
Thx for you contribution. 
Unfortunately I have to pas on this. 
This change already landed in an other PR.

PR #261, opened 2026-06-23
fix: preserve functional data (URLs, paths, commands) from shrink

pkking:
## Problem

The `shrink` tag in ponytail-review treated URLs as decoration and stripped protocol prefixes (`https://`), breaking clickability in editors/terminals.

Example: A 7-line comment with full URLs was shrunk to 2 lines with incomplete URLs like `alibabacloud.com/help/...` instead of `https://www.alibabacloud.com/help/...`.

## Root Cause

The `shrink` tag definition said "same logic, fewer lines" but didn't distinguish between:
- **Shrinkable code logic** (can be compressed if functionally equivalent)
- **Non-shrinkable functional data** (URLs, file paths, commands that must remain complete to work)

## Solution

Added explicit rules to both `ponytail` and `ponytail-review` skills:

### ponytail/SKILL.md
Added to "When NOT to be lazy" section:
```
**Never shrink functional data.** URLs must include protocols (`https://`),
file paths must remain complete, commands must stay executable. A URL
without its protocol isn't "the same logic, shorter" — it's a broken URL.
Functional strings are executable data, not decoration.
```

### ponytail-review/SKILL.md
Added to "Boundaries" section:
```
**Never shrink functional data.** URLs must include protocols (`https://`),
file paths must remain complete and valid, commands must stay executable.
A URL without its protocol isn't "the same logic, shorter" — it's a broken
URL. Functional strings are executable data, not decoration.
```

### examples/functional-data.md
Added example demonstrating correct behavior:
- Shows broken case (URLs without protocols)
- Shows correct case (full URLs)
- Explains the principle

### test/functional-data.test.js
Added test suite verifying:
- ponytail SKILL.md contains functional data rule
- ponytail-review SKILL.md contains functional data rule
- functional data example exists and demonstrates the principle

All 3 tests pass.

## Changes

- `skills/ponytail/SKILL.md`: Added rule to "When NOT to be lazy" section
- `skills/ponytail-review/SKILL.md`: Added rule to "Boundaries" section
- `examples/functional-data.md`: New example file (correct behavior)
- `test/functional-data.test.js`: New test file (3 tests, all passing)

## Impact

This prevents future instances of ponytail-review suggesting to shrink functional data (URLs, paths, commands) in the name of brevity, while preserving the core principle of aggressive code simplification.

Net: +90 lines (documentation + tests), but prevents broken URLs and improves clarity about what can/cannot be shrunk.


pkking:
@DietrichGebert could you please take a look this PR :)
I appreciated ponytail, and want to help

DietrichGebert (commented):
Thanks for this, the rule itself is correct. The problem is that it only landed in 2 of the files that carry this content, so the other integrations are now out of parity and the test suite goes red. I checked the branch out and ran it to confirm.

## Blocking: `npm test` fails (4 new failures)

The canonical skill bodies changed, but the generated OpenClaw copies were not regenerated, so `tests/openclaw-skills.test.js` fails:

```
x ponytail: committed OpenClaw skill matches the generator
x ponytail: body is the canonical skills/ponytail body, verbatim
x ponytail-review: committed OpenClaw skill matches the generator
x ponytail-review: body is the canonical skills/ponytail-review body, verbatim
```

`main` is 54 pass / 2 fail, this branch is 50 pass / 6 fail. (The 2 pre-existing email and csv failures are unrelated to this PR.)

Fix: run `node scripts/build-openclaw-skills.js` and commit the regenerated `.openclaw/skills/ponytail/SKILL.md` and `.openclaw/skills/ponytail-review/SKILL.md`.

## Blocking: the new tests never run

`package.json` runs `node --test tests/*.test.js` (plural `tests/`), but the new file is at `test/functional-data.test.js` (singular `test/`). `npm test` does not pick it up, so the 3 tests only pass when run by hand. Please move it to `tests/functional-data.test.js`.

## Parity: the rule is missing from the rest of the surface

The `shrink:` tag and the "When NOT to be lazy" carve-out are duplicated across the integration files. The fix currently reaches only `skills/ponytail/SKILL.md` and `skills/ponytail-review/SKILL.md`. The same content still lives in:

`shrink:` tag (the exact "same logic, fewer lines" text this PR identifies as the root cause):
- `skills/ponytail-audit/SKILL.md:23` (audit defines the identical tag, still unguarded)
- `commands/ponytail-review.toml:2` and `commands/ponytail-audit.toml:2`
- `.opencode/command/ponytail-review.md:5` and `.opencode/command/ponytail-audit.md:5`

"Not lazy about" carve-out (home of the base rule):
- `AGENTS.md:30`, then re-sync its 6 byte-equal mirrors so `check-rule-copies.js` stays green: `.cursor/rules/ponytail.mdc`, `.windsurf/rules/ponytail.md`, `.clinerules/ponytail.md`, `.agents/rules/ponytail.md`, `.github/copilot-instructions.md`, `.kiro/steering/ponytail.md`
- optional: the hardcoded fallback string in `hooks/ponytail-instructions.js`

The Claude hook, Pi extension, and MCP server read `skills/ponytail/SKILL.md` at runtime, so those need no change.

## Suggested order

1. Add the rule to `skills/ponytail-audit/SKILL.md` and `AGENTS.md`, then propagate the `AGENTS.md` body into the 6 mirrors.
2. Tighten the `shrink` wording in the 4 command prompts.
3. Run `node scripts/build-openclaw-skills.js` to refresh every OpenClaw copy.
4. Move the test into `tests/`.
5. `npm test` should be back to 54 pass / 2 fail.

pkking:
@DietrichGebert fixed :)

PR #269, opened 2026-06-23
Add Claude Code .md slash commands so /ponytail* works in Claude Code

jjisnow:
## Problem

Installed ponytail into the **Claude Code extension for VS Code** (v4.7.0) and the `/ponytail*` slash commands never showed up. The skills (`skills/*/SKILL.md`) and hooks load fine — it's specifically the slash commands that are missing.

Root cause: Claude Code only discovers slash commands from **`commands/*.md`** (Markdown + YAML frontmatter). The repo's `commands/*.toml` files are the Codex/opencode/Gemini format, which Claude Code doesn't parse — so `/ponytail`, `/ponytail-review`, `/ponytail-audit`, `/ponytail-debt`, `/ponytail-gain`, and `/ponytail-help` never register there. (This is a Claude Code behavior, not a VS Code one — it'll be the same in the CLI; VS Code is just where I hit it.)

## Change

Add six `commands/*.md` files mirroring the existing `.toml` set 1:1 — same `description` and prompt for each, with `{{args}}` → `$ARGUMENTS` for the level switch. Purely additive: the skills, hooks, and `.toml` commands are untouched, so nothing changes for Codex/opencode/Gemini users.

After this, the commands appear in Claude Code's `/` menu and run the same prompts they do elsewhere.

## Notes

- Kept it minimal and in the spirit of the project — thin wrappers, no new docs or deps.
- Verified locally: dropping these `.md` files in makes the commands resolve in Claude Code.
- Separately, I hit a broken install where the plugin cache dir was empty (no files copied), which also blocks the skills/hooks — but that looked like a local install glitch (Claude Code extension in VS Code) rather than a repo issue, so it's out of scope for this PR.

### DIFF
--- a/commands/ponytail-audit.md
+++ b/commands/ponytail-audit.md
@@ -1,0 +1,5 @@
+---
+description: Audit the whole repo for over-engineering, what can be deleted
+---
+
+Audit the entire repository for over-engineering only, not correctness. Scan the whole tree, not a diff. One line per finding, ranked biggest cut first: <tag> <what to cut>. <replacement>. [path]. Tags: delete (dead code/speculative feature), stdlib (reinvented standard library), native (dependency doing what the platform does), yagni (abstraction with one implementation), shrink (same logic, fewer lines). End with the net lines and dependencies removable. If nothing to cut: 'Lean already. Ship.'
--- a/commands/ponytail-debt.md
+++ b/commands/ponytail-debt.md
@@ -1,0 +1,5 @@
+---
+description: Harvest ponytail: comments into a tracked debt ledger
+---
+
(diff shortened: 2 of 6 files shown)


DietrichGebert:
Thanks for the careful writeup, and you nailed the real bug: `commands/*.toml` is the Gemini CLI format, not Claude Code's. That mislabel is on me, the test comment in `commands.test.js` literally calls it "Claude Code", which is wrong.

Here's why I'm going to pass on this one though. On current Claude Code builds the plugin's skills already register as slash commands: `/ponytail`, `/ponytail-review`, `/ponytail-audit`, `/ponytail-debt`, `/ponytail-gain`, and `/ponytail-help` all show in the `/` menu with their descriptions, and `/ponytail` even shows the `[lite|full|ultra]` arg hint. And the mode switch doesn't run through a command at all: the `UserPromptSubmit` hook (`ponytail-mode-tracker.js`) parses the prompt text, so `/ponytail ultra` persists the mode whether it came from a skill or was typed raw. It even handles the `/ponytail:ponytail` form.

So adding `commands/*.md` on top would just duplicate those menu entries rather than add behavior. We're both on ponytail 4.7.0 and the skills show as slash commands on my end, so this isn't a repo gap, it's the Claude Code build you were on not surfacing skills that way yet (or the empty-cache install you flagged). The hooks still load regardless, so `/ponytail ultra` typed as text works even without the menu entry.

Appreciate the diagnosis and the clean additive approach. Closing this one, but the toml-is-not-Claude catch is real and I'll fix that misleading test comment separately.

PR #287, opened 2026-06-24
fix(hooks): interpolate settingsPath in statusline nudge (closes #255)

### LINKED ISSUE #255: Statusline setup nudge hardcodes ~/.claude/settings.json, ignoring CLAUDE_CONFIG_DIR
## Problem

`ponytail-activate.js`'s "STATUSLINE SETUP NEEDED" nudge hardcodes `~/.claude/settings.json` as the file to edit. When `CLAUDE_CONFIG_DIR` is set, the flag file and settings live under that directory instead — so the nudge sends the user to the wrong file, and the statusline never gets configured.

This is the same `CLAUDE_CONFIG_DIR` override already honored for the `.ponytail-active` flag (issue #34); the nudge text just didn't follow suit.

## Fix

Use the already-resolved `settingsPath` (computed from `getClaudeDir()`, which honors `CLAUDE_CONFIG_DIR`) in both nudge branches — the shell-safe snippet and the manual-config fallback — instead of the literal `~/.claude/settings.json`.

## Scope

- `hooks/ponytail-activate.js` — two string substitutions
- `tests/hooks.test.js` — assert the nudge references `CLAUDE_CONFIG_DIR/settings.json` and not `~/.claude/settings.json`

Tanmay9223:
🎯 **What:** Replaces the two hardcoded `~/.claude/settings.json` strings in the statusline setup nudge with the already-computed `settingsPath` variable, which correctly resolves `CLAUDE_CONFIG_DIR`.

💡 **Why:** When `CLAUDE_CONFIG_DIR` is set to a custom location, `ponytail-activate.js` already reads the flag file and detects the missing statusline from the correct directory — but then told the user (and the agent) to edit `~/.claude/settings.json`, which Claude Code is not reading. This caused the statusline to silently never activate for any user with a custom config dir.

✅ **Verification:** Extended the existing `CLAUDE_CONFIG_DIR` test block in `tests/hooks.test.js` with two new assertions: one that confirms the resolved `settingsPath` appears in the nudge output, and one that confirms the hardcoded `~/.claude/settings.json` literal does not. All 60 of 61 tests pass; the single remaining failure is the pre-existing `csv: correct pandas one-liner passes` test which requires `pandas` and was already failing on `main` before this change.

✨ **Result:** Users with `CLAUDE_CONFIG_DIR` set now receive a nudge pointing at the correct settings file, so the statusline can actually be configured and activated.

### DIFF
--- a/hooks/ponytail-activate.js
+++ b/hooks/ponytail-activate.js
@@ -66,7 +66,7 @@ if (!isCodex && !isCopilot) try {
       output += "\n\n" +
         "STATUSLINE SETUP NEEDED: The ponytail plugin includes a statusline badge showing active mode " +
         "(e.g. [PONYTAIL], [PONYTAIL:ULTRA]). It is not configured yet. " +
-        "To enable, add this to ~/.claude/settings.json: " +
+        `To enable, add this to ${settingsPath}: ` +
         statusLineSnippet + " " +
         "Proactively offer to set this up for the user on first interaction.";
     } else {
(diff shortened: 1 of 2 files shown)


Tanmay9223:
Closing this PR because the original issue it addresses (#255) has been closed.

PR #291, opened 2026-06-24
fix: read AGENTS.md as fallback instead of hardcoded duplicate

Parithosh-Varma:
The `getFallbackInstructions` in `hooks/ponytail-instructions.js` hardcoded a 33-line rule duplicate that silently drifted from the canonical `AGENTS.md`. It had extra sections (Persistence, Output, Boundaries), different rule formatting, and was missing rules present in AGENTS.md. This meant that if `skills/ponytail/SKILL.md` was ever unreadable, the agent would receive stale instructions with no indication anything was wrong.

The fix reads `AGENTS.md` directly as the primary fallback source. If both `SKILL.md` and `AGENTS.md` are unavailable, only then does it fall back to a minimal ladder (7 lines vs the previous 33). This eliminates the drift problem entirely since the fallback always mirrors the canonical rules.


DietrichGebert:
Thanks for this. Reading AGENTS.md instead of keeping a second copy is the right call. I am closing it in favor of #303, which makes the same change and also turns those silent catch blocks into real stderr logs in one go. Thanks again for digging in.

PR #301, opened 2026-06-24
fix(opencode): stop exporting parseCommandFile from plugin module

ayushere:
OpenCode's legacy plugin loader iterates every exported function in a plugin module and tries to invoke each one with the plugin context object. `parseCommandFile` was exported alongside the default plugin function, so the loader called it with an object instead of a file path and crashed with *"path must be a string or a file descriptor"* before any hook ran, which aborted the whole plugin load (v4.8.3, observed on OpenCode 1.17.9).

## Repro

```json
// ~/.config/opencode/opencode.jsonc
{ "plugin": ["@dietrichgebert/ponytail"] }
```

```
opencode run "test" --print-logs
# ERROR  failed to load plugin  path=@dietrichgebert/ponytail
#        error="path must be a string or a file descriptor"
```

No hooks fire, no commands register, the plugin is silently dead.

## Fix

Move the frontmatter parser out of the plugin module into its own .cjs sibling. The plugin module now exposes exactly one top-level export (the default plugin function); the loader has nothing else to mis-invoke.

- `.opencode/plugins/ponytail-frontmatter.cjs` (new) — the parser, same regex, same CRLF tolerance
- `.opencode/plugins/ponytail.mjs` — drops the named `parseCommandFile` export, requires the sibling via the existing `createRequire` bridge
- `tests/opencode-plugin.test.js` — requires the sibling directly; existing frontmatter tests are unchanged in intent

## Verification

```
node --test tests/opencode-plugin.test.js
# 6 pass, 0 fail

node -e "import('./.opencode/plugins/ponytail.mjs').then(m => console.log(Object.keys(m)))"
# [ 'default' ]
```

(`npm test`'s CSV row fails on this machine because pandas isn't installed; the same test fails on `main` without these changes, so it's pre-existing and unrelated.)

### DIFF
--- a/.opencode/plugins/ponytail-frontmatter.cjs
+++ b/.opencode/plugins/ponytail-frontmatter.cjs
@@ -1,0 +1,23 @@
+'use strict';
+
+// ponytail command-file frontmatter parser.
+//
+// Pulled out of ponytail.mjs so the plugin module's only top-level export is
+// the plugin function itself. OpenCode's legacy plugin loader (the one that
+// runs before v1 plugins are detected) treats every function exported from a
+// plugin module as a plugin; calling the frontmatter parser as one threw
+// "path must be a string or a file descriptor" because it got the plugin
+// context object as its first argument. Keeping the parser in its own module
+// leaves exactly one plugin-shaped export on ponytail.mjs.
+
+function parseCommandFile(filePath) {
+  const fs = require('fs');
+  const content = fs.readFileSync(filePath, 'utf8');
+  // Tolerate CRLF: a Windows checkout (autocrlf) delivers \r\n, npm ships \n.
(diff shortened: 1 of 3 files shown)


spde:
Confirming this independently — installing the OpenCode plugin (v4.8.x) fails to load with `error="path must be a string or a file descriptor"` and silently disables all hooks/commands. Same root cause as described here: OpenCode's loader invokes every exported function with the plugin context object, so the named `parseCommandFile` export gets called as `parseCommandFile(ctx)` → `fs.readFileSync(<object>)`.

Verified that dropping the export (this PR) makes the plugin load cleanly again on OpenCode 1.17.x. This silently broke a plugin that measurably reduced token usage in our setup, so a merge would be much appreciated 🙏

PR #303, opened 2026-06-25
fix: log errors to stderr instead of swallowing silently

Parithosh-Varma:
Every try/catch across hooks and the OpenCode plugin swallowed errors silently with empty bodies or comments. When a hook failed (SKILL.md unreadable, JSON parse error, filesystem issue, EPIPE), there was zero evidence in any log — debugging required source patches to find out what happened.

Now each catch writes the error message to stderr before continuing with the same graceful fallthrough behavior. Production hooks still never crash the agent; operators can now see what failed.

Affected files:
- `hooks/ponytail-activate.js` (3 catches: setMode, statusline detection, writeHookOutput)
- `hooks/ponytail-subagent.js` (1 catch: writeHookOutput at exit)
- `hooks/ponytail-config.js` (1 catch: config file read)
- `hooks/ponytail-runtime.js` (1 catch: flag file deletion)
- `hooks/ponytail-mode-tracker.js` (1 catch: JSON parse error)
- `.opencode/plugins/ponytail.mjs` (2 catches: log call, commandDir read)

### DIFF
--- a/.openclaw/skills/ponytail-help/SKILL.md
+++ b/.openclaw/skills/ponytail-help/SKILL.md
@@ -26,6 +26,8 @@ Level sticks until changed or session end.
 |-------|---------|--------------|
 | **ponytail** | `/ponytail` | Lazy mode itself. Simplest solution that works. |
 | **ponytail-review** | `/ponytail-review` | Over-engineering review: `L42: yagni: factory, one product. Inline.` |
+| **ponytail-audit** | `/ponytail-audit` | Whole-repo over-engineering audit, ranked by impact. |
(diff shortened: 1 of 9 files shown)


DietrichGebert:
Thanks for this. A few of the silent catches it touches are actually deliberate, they carry comments about not surfacing a stdout/EPIPE close at hook exit as a hook failure, and writing to stderr there can re-introduce exactly that. On top of that the help-card rows are already on main via #480, and the getFallbackInstructions rewrite to read AGENTS.md is a real change to the injected ruleset that we'd want to benchmark on its own rather than land under a "log to stderr" PR. Going to close this. If there's one specific spot swallowing an error users actually need to see, a small targeted PR for that would be great. Thanks!

PR #304, opened 2026-06-25
fix: keep OpenCode system prompt Qwen-compatible

### LINKED ISSUE #296: OpenCode: BadRequestError on Qwen models
Hi,
Thanks for this tooling.
Hope to help make it a bit better :)

Agent: OpenCode
Model: Qwen/Qwen35-397B-A17B-FP8 (same issue with other Qwens)

Scenario: 
1. Modify OpenCode config `{ "plugin": ["@dietrichgebert/ponytail"] }`
2. Run OpenCode
3. Send a single request

Expected result:
No error

Actual result:
```
Bad Request: {"detail":"{\"error\":{\"message\":\"System message must be at the beginning.\",\"type\":\"BadRequestError\",\"param\":null,\"code\":400}}"}
```

Also:
The issue cannot be reproduced with non-Qwen models like Gemma, gpt-oss.

snvtac:
## Summary
- Merge Ponytail's OpenCode instructions into the existing system prompt instead of appending a second system message.
- Preserve the no-baseline fallback and `/ponytail off` behavior.
- Add an OpenCode regression test that keeps the system array at one entry when OpenCode supplies a baseline prompt.

Closes #296

## Testing
- `node --test tests/opencode-plugin.test.js`
- `node scripts/check-versions.js`
- `git diff --check`
- `PATH="$PWD/.venv/bin:$PATH" npm test` (with pandas installed in a repo-local uv venv)

### DIFF
--- a/.opencode/plugins/ponytail.mjs
+++ b/.opencode/plugins/ponytail.mjs
@@ -82,7 +82,12 @@ export default async ({ client } = {}) => {
     'experimental.chat.system.transform': async (_input, output) => {
       const mode = readMode();
       if (mode === 'off') return;
-      output.system.push(getPonytailInstructions(mode));
+      const instructions = getPonytailInstructions(mode);
+      if (output.system.length) {
+        output.system[0] = `${output.system[0]}\n\n${instructions}`;
+      } else {
+        output.system.push(instructions);
+      }
     },
 
     // Persist `/ponytail <level>` so the next turn's injection follows it.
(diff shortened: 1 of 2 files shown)


Terminator637:
Any updates there? Still can't use Qwen models with OpenCode

kevinnio:
@DietrichGebert — could you take a look at this PR when you have a moment? Thanks!

DietrichGebert:
Thanks for the Qwen compatibility fix. #536 handles the same OpenCode system entry merge and links the tracking issue, so I am going with that one. Thanks for flagging the problem.

PR #320, opened 2026-06-25
Add external DeepSWE/Pi benchmark results

Whamp:
Adds an external DeepSWE/Pi benchmark writeup.

This is deliberately framed as a stress test, not a replacement for the existing Ponytail benchmark. Result is mixed:

- patch lines: -26%
- tokens: -12%
- cost: -10%
- time: -7%
- full solves: 2/113 → 4/113
- mean partial reward: 0.774 → 0.709

The interesting part is the mechanical analysis: Ponytail helped by pruning work in some cases, but easy/medium regressions were often over-minimization — missed tests, fixtures, exports, generated files, or downstream wiring.

Use it, edit it, or reject it freely. I thought the negative result was useful for the project because it clarifies where “lazy” becomes “too lazy.”

Source harness/results: https://github.com/Whamp/deep-swe-bench

Validation: docs-only change; checked markdown diff with `git diff --check`.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -70,6 +70,10 @@ Versus baseline, ponytail writes **80-94% less code**, costs **42-75% less**, an
 > while the bare "one-liner" prompt drops a guard. That is the honest, defensible number. See
 > [results/2026-06-18-agentic.md](results/2026-06-18-agentic.md).
 
+## Additional results
+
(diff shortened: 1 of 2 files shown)


DietrichGebert:
Thanks for this, Will, and especially for sharing a negative result. That kind of writeup is exactly the sort of thing the project benefits from.

Before any benchmark numbers land in the repo I need to run and verify them myself, so I'm going to close this for now while I do that. Nothing wrong with the PR itself. I just want to reproduce the DeepSWE/Pi results on my end before they get published as part of the benchmarks.

I'll follow up once I've checked them. Thanks again for the harness and results link.

PR #328, opened 2026-06-26
fix: add pi status hide config

### LINKED ISSUE #324: Feature: hide the status bar indicator via config while keeping ponytail active
## Summary

I use ponytail via pi. The pi extension draws a status bar indicator
(`○ 🐴 ponytail: ⚡ FULL`) whenever a mode is active, with no way to suppress just
the indicator while keeping the lazy-dev ruleset injection on. The only way to
blank the line is `/ponytail off`, which also stops the ruleset each turn.

## Problem or pain point

The indicator is drawn unconditionally in `syncStatus()` (`pi-extension/index.js`):
whenever `currentMode !== "off"` it calls `c.ui.setStatus(...)`. Only the `off`
mode avoids drawing it, and that also disables the ruleset.

The shared config resolver (`hooks/ponytail-config.js`) only supports
`defaultMode`, via `PONYTAIL_DEFAULT_MODE` or the `defaultMode` field in
`~/.config/ponytail/config.json`. There is no `hideStatus` flag.

Users who want the always-on ruleset but a clean status bar have no option except
turning ponytail off entirely, which removes the behavior they want.

## Proposed behavior

Add a config option to skip the indicator independently of mode, mirroring the
existing `defaultMode` pattern:

- env: `PONYTAIL_HIDE_STATUS=1`
- config: `hideStatus: true` in `~/.config/ponytail/config.json`

When set, `syncStatus()` returns early without calling `c.ui.setStatus`, while
`currentMode` and the `before_agent_start` ruleset injection stay active.

## Alternatives considered

- `/ponytail off`: clears the indicator but also disables ruleset injection.
- The indicator is drawn by the extension itself, so it cannot be hidden from
  host settings without a ponytail-side flag.

## Additional context

Verified against v4.8.3 (latest release) and `main`: `syncStatus()` on `main`
matches the installed copy and has no hide flag.

sanmaxdev:
## Summary
- Add `PONYTAIL_HIDE_STATUS=1` and `hideStatus: true` config support for the Pi status indicator.
- Keep mode tracking and instruction injection active when the indicator is hidden.
- Preserve `hideStatus` when updating the saved default mode.

## Testing
- [x] `npm test --prefix pi-extension`
- [x] `npm test`
- [x] `node scripts/check-rule-copies.js`
- [x] `node scripts/check-versions.js`
- [x] `git diff --check`

Closes #324

### DIFF
--- a/hooks/ponytail-config.js
+++ b/hooks/ponytail-config.js
@@ -8,6 +8,8 @@
 //      - ~/.config/ponytail/config.json (macOS / Linux fallback)
 //      - %APPDATA%\ponytail\config.json (Windows fallback)
 //   3. 'full'
+//
+// Status indicator hiding uses PONYTAIL_HIDE_STATUS=1 or config.hideStatus=true.
 
 const fs = require('fs');
 const path = require('path');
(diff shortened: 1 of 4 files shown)


DietrichGebert:
Closing in favor of #544 (merged), which reconciles this PR and #511 and credits you both as co-authors. It keeps your caching + session_start re-read and the stronger test (status hidden while the ruleset stays active), adds a BOM-safe config read, and drops the writeDefaultMode change here since that already landed in #514. Thanks for the feature.

PR #330, opened 2026-06-26
feat: add Kiro global install script

mccannical:
Adds `scripts/install-kiro.sh` — a one-command installer for the Kiro IDE.

## What it does

Symlinks `.kiro/steering/ponytail.md` from a local clone into `~/.kiro/steering/` so the ruleset loads globally in every Kiro session.

- Idempotent (re-running is a no-op if already linked)
- Updates come from `git pull` — no stale copies
- Uninstall: `rm ~/.kiro/steering/ponytail.md`

## Usage

```bash
git clone https://github.com/DietrichGebert/ponytail.git
./ponytail/scripts/install-kiro.sh
```

Follows the same pattern as the existing Kiro guidance in the README (copy to `~/.kiro/steering/`) but automates it with a symlink for easier maintenance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,15 +241,22 @@ clawhub install ponytail
 
 Installs ponytail as an OpenClaw skill from ClawHub; the review, audit, debt, gain, and help skills install the same way (`clawhub install ponytail-review`, and so on). OpenClaw applies it on coding tasks and also exposes it as a `/ponytail` command. Without ClawHub, copy [`.openclaw/skills/ponytail`](.openclaw/skills/) into `~/.openclaw/skills/`.
 
+### Kiro
+
+```bash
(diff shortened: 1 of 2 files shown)


DietrichGebert:
Thanks for this, and the script itself is clean (idempotent, good error handling). It goes against how ponytail handles rule-only hosts though: every one of them, Kiro included, just says "copy the file", one line, cross-platform, nothing to maintain. A bespoke installer for a single host is a first here, and being bash-only it actually drops the Windows-friendly global-copy instruction Kiro users have today, and Kiro runs on Windows too. Anyone who wants the auto-updating symlink can run a one-line ln -s themselves. So I am going to keep the existing one-line copy instruction and close this. The uninstall-row idea was a nice touch though, thanks for the effort.

PR #337, opened 2026-06-27
Fix retry loop & hook timeouts

Danesh-Ali:
### Problem

Two issues were causing excessive API retries and socket drops during longer sessions:

1. **`pre_llm_call` fired on every LLM turn** - ~5KB of Ponytail context was injected before every single API call. In long conversations this ballooned context size continuously, eventually triggering `The socket connection was closed unexpectedly` errors and retry loops.

2. **Hook timeout too low (5s)** - On slow machines or cold `node` starts, hooks timed out before completing, causing Claude to retry session initialization.

### Changes

**`__init__.py`**
- Added `_injected_sessions: set[str]` to track which sessions have already received context.
- `_pre_llm_call` now returns `None` for any `session_id` seen before - context injected once per session, not once per turn.

**`hooks/claude-codex-hooks.json`**
- Bumped `timeout` from `5` to `10` across all three hooks (`SessionStart`, `SubagentStart`, `UserPromptSubmit`).

### Result

- Context injection: **once per session** instead of every turn.
- Hook timeout headroom doubled - eliminates cold-start failures on slow machines.
- Eliminates the primary causes of socket drops and retry spirals in agentic workflows.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -122,10 +122,18 @@ def build_injected_context(mode: str | None = None) -> str:
         return _fallback_instructions(effective)
 
 
+_injected_sessions: set[str] = set()
+
 def _pre_llm_call(session_id: str = "", **_: Any) -> dict[str, str] | None:
+    if session_id and session_id in _injected_sessions:
+        return None
     mode = _current_mode or _default_mode()
     context = build_injected_context(mode)
-    return {"context": context} if context else None
+    if context:
+        if session_id:
+            _injected_sessions.add(session_id)
+        return {"context": context}
+    return None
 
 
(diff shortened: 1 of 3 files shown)


DietrichGebert:
Thanks for digging into the hook timeout stuff. Since you opened this, main rewrote both the subagent hook (#522) and the codex hook command (#477), so two of the three changes here target code that no longer exists, and the retry loop it fixes isn't present on current main. The one new bit, the per-session injection dedup in __init__.py, would change how often the rules get injected, which we'd need to benchmark on its own. Going to close this against current main. If you still hit a real timeout on latest, please reopen with a repro. Thanks!

PR #340, opened 2026-06-27
fix(benchmarks): match CRLF code fences in loc.js

### LINKED ISSUE #339: benchmarks: loc.js misses CRLF code fences and counts the whole response

```js
/```[a-zA-Z0-9_+-]*
([\s\S]*?)```/g
```

There is no `
?` before the `
`, so a CRLF fence (` ```js

?

Repro:

```js
const lf   = 'Here:
```js
const a = 1;
const b = 2;
```
Done.';
console.log(loc(lf.replace(/
/g, '
```

### PR DESCRIPTION
**Pull request by Lakshya77089**
Fixes #339.

`loc.js` extracts code blocks with `/```[a-zA-Z0-9_+-]*
([\s\S]*?)```/g` — no `
?` before the `
`, so a CRLF fence (` ```js
 `) never matches. With no block found it falls back to counting the entire response, prose and fence lines included.

`correctness.js` already handles this with `
?
`, so on CRLF input the LOC metric and the correctness gate disagree. This aligns `loc.js` with it.

Before:
```
loc(lf)              -> 2   (correct)
loc(lf as CRLF)      -> 6   (counts prose + fences)
```
After: both score 2.

Added a CRLF case to `benchmarks/loc.test.js` (now 6/6). LF input is unaffected; one-token regex change.

### DIFF
--- a/benchmarks/loc.js
+++ b/benchmarks/loc.js
@@ -3,7 +3,9 @@
 // Recorded as the `code_loc` metric per arm (always passes; it is a measurement, not a gate).
 module.exports = (output) => {
   const text = String(output || '');
-  const blocks = [...text.matchAll(/```[a-zA-Z0-9_+-]*\n([\s\S]*?)```/g)].map((m) => m[1]);
+  // \r? so CRLF fences match too; without it, CRLF output finds no block and
+  // the whole response (prose included) gets counted. Mirrors correctness.js.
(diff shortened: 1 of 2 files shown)


DietrichGebert:
Closing: #339 was fixed in #518 (merged). Your test (prose-wrapped CRLF fence) was nicer, but #518 also removed the dead codeOf function, which resolves the companion #533 in the same change. Thanks for the fix and the clear repro.

Issue #260, opened 2026-06-23
Inconsistent versioning


CamxxCore:
The root package.json has version 0.1.0, while Claude/Codex/Gemini manifests say 4.7.0, and GitHub’s latest release is v4.8.0.

Probably worth fixing and pinning what is actually being served, for clarity. Awesome and promising looking project btw 👍

gajicm93:
+1, Claude reports 4.7.0 as latest.

DietrichGebert:
Fixed in v4.8.1: https://github.com/DietrichGebert/ponytail/releases/tag/v4.8.1

You were right. All six version-bearing files are now aligned to 4.8.1: the four plugin manifests (Claude, Codex, Copilot, Gemini) plus the root package.json and ponytail-mcp/package.json. v4.8.0 was already tagged with the stale manifests, so the fix shipped as a fresh v4.8.1 rather than rewriting a published tag.

To keep it from drifting again, there is now a CI guard (scripts/check-versions.js) that fails the build when the version files disagree, or when a release tag does not match the manifest version.

Thanks for flagging it, and for the kind words.

PR #357, opened 2026-06-27
fix: add plugin.yaml and pi-extension/package.json to VERSION_FILES

### LINKED ISSUE #348: check-versions.js: plugin.yaml and pi-extension/package.json missing from VERSION_FILES
In `scripts/check-versions.js`, the `VERSION_FILES` array does not include `plugin.yaml` or `pi-extension/package.json`. This means version drift in these files won't be caught — the exact class of bug that issues #260/#262 were filed for.

Add both files to `VERSION_FILES` to prevent future drift.

Affected: `scripts/check-versions.js`

nanaubusiness:
These two version-bearing files were missing from the `check-versions.js` consistency check, creating the exact drift scenario that issues #260/#262 were filed for.

Changes:
- Add `plugin.yaml` (Hermes manifest) to `VERSION_FILES`
- Add `pi-extension/package.json` to `VERSION_FILES`
- Update comment count from 7 to 9 files

Closes #348

### DIFF
--- a/scripts/check-versions.js
+++ b/scripts/check-versions.js
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
-// Version-consistency guard. Ponytail declares its version in seven files across
-// five host ecosystems, and every release bumps all of them by hand.
+// Version-consistency guard. Ponytail declares its version in nine files across
+// six host ecosystems, and every release bumps all of them by hand.
 //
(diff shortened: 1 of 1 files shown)


DietrichGebert:
Closing: #348 has since been closed as resolved, so this PR is no longer needed. Thanks for the contribution — if you think the fix didn't fully land, comment and we'll re-open.

PR #359, opened 2026-06-27
fix: correct relative import paths for npm-installed opencode plugin

nanaubusiness:
In `.opencode/plugins/ponytail.mjs`, the require paths used `../../hooks/` which resolves outside the package root when installed via npm. Changed to `../hooks/` which works in both checkout and npm-install contexts.


DietrichGebert:
Thanks for digging into this. The current ../../hooks path already resolves fine, both in the repo and from an npm install, because hooks/ sits at the package root, and ../hooks would point at .opencode/hooks which does not exist. So this one would actually break the plugin. I am going to close it, but I appreciate the report.

PR #367, opened 2026-06-27
fix: handle /ponytail:review (colon form) for review mode

### LINKED ISSUE #361: hooks/ponytail-mode-tracker.js: /ponytail:review silently activates wrong mode
In `hooks/ponytail-mode-tracker.js`, when a user types `/ponytail:review` (colon-separated form), it is silently treated as the default mode ('full') instead of activating 'review' mode.

Trace:
- `prompt = '/ponytail:review'`
- Regex `/^[/@$]ponytail/` matches
- `parts = ['/ponytail:review']` (one element, no space to split on)
- `cmd = '/ponytail:review'`
- `cmd === '/ponytail-review'` → FALSE
- Falls to `mode = getDefaultMode()` → **'full'**

The user explicitly typed review mode but gets full mode with no warning.

nanaubusiness:
Previously '/ponytail:review' fell through to getDefaultMode() instead of activating review mode — the condition checked for '/ponytail:ponytail-review' (wrong pattern, also unreachable). Now correctly matches '/ponytail:review'.

Also removes the unreachable `/ponytail:ponytail-review` disjunct that could never be matched.

Closes #361

### DIFF
--- a/hooks/ponytail-mode-tracker.js
+++ b/hooks/ponytail-mode-tracker.js
@@ -21,7 +21,7 @@ process.stdin.on('end', () => {
 
       let mode = null;
 
-      if (cmd === '/ponytail-review' || cmd === '/ponytail:ponytail-review') {
+      if (cmd === '/ponytail-review' || cmd === '/ponytail:review') {
         mode = 'review';
       } else if (cmd === '/ponytail' || cmd === '/ponytail:ponytail') {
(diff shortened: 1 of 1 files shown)


DietrichGebert:
Closing: #361 has since been closed as resolved, so this PR is no longer needed. Thanks for the contribution — if you think the fix didn't fully land, comment and we'll re-open.

PR #368, opened 2026-06-27
fix: correct CSV sum regex to use \d instead of [\\d]

### LINKED ISSUE #365: benchmarks/correctness.js: regex [\\d] matches literal backslash/d not digit
In `benchmarks/correctness.js` line 207:

```python
if re.search(r'(?<![\\d])351(?:\\.0)?(?![\\d])', output):
```

In a Python raw string, `[\\d]` is the character class containing a literal backslash and the letter 'd', NOT the digit shorthand `\d`. The intent was `(?<!\d)` / `(?!\d)` (not preceded/followed by a digit), but the double-backslash makes it check for backslash or 'd' instead.

This would cause the correctness gate to mis-evaluate test cases involving the number 351.

nanaubusiness:
In `benchmarks/correctness.js`, the character class `[\\d]` in a Python raw string matches a literal backslash or 'd', not a digit. The intended lookbehind/lookahead `(?<!\d)` and `(?!\d)` now work correctly to detect the standalone number 351 in CSV sum output.

Closes #365

### DIFF
--- a/benchmarks/correctness.js
+++ b/benchmarks/correctness.js
@@ -204,7 +204,7 @@ sys.stdout = _stdout
 # Check output contains the number 351 (100.5 + 200.0 + 50.5)
 # Match as a standalone number (not as substring of e.g. 13510)
 import re
-if re.search(r'(?<![\\d])351(?:\\.0)?(?![\\d])', output):
+if re.search(r'(?<!\d)351(?:\.0)?(?!\d)', output):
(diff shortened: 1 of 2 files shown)


DietrichGebert:
Closing: #365 has since been closed as resolved, so this PR is no longer needed. Thanks for the contribution — if you think the fix didn't fully land, comment and we'll re-open.

PR #369, opened 2026-06-27
fix: remove false claim that pi-extension exposes ponytail: skill aliases

### LINKED ISSUE #364: after-install.md: promises skill aliases that pi-extension never registers
In `after-install.md` lines 22-23:

```
Bundled skills are available as `ponytail:ponytail`, `ponytail:ponytail-review`, ...
```

But `pi-extension/index.js` only calls `pi.registerCommand()` for slash commands. There is **no skill registration API call** that would expose `ponytail:<skill>` aliases. A user reading the docs and trying to use `ponytail:ponytail` with the pi-extension will get an error — the feature simply doesn't exist.

nanaubusiness:
The pi-extension only registers slash commands (e.g. /ponytail), not skill-style aliases (ponytail:ponytail). The claim in after-install.md was inaccurate.

Closes #364

### DIFF
--- a/after-install.md
+++ b/after-install.md
@@ -19,4 +19,4 @@ Commands:
 - `/ponytail-gain`
 - `/ponytail-help`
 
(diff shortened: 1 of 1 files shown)


DietrichGebert:
Closing: #364 has since been closed as resolved, so this PR is no longer needed. Thanks for the contribution — if you think the fix didn't fully land, comment and we'll re-open.

PR #371, opened 2026-06-27
fix: use hookSpecificOutput JSON for UserPromptSubmit mode changes

### LINKED ISSUE #363: hooks/ponytail-runtime.js: UserPromptSubmit mode-change message replaces user prompt
In `hooks/ponytail-runtime.js` lines 53-58, for non-SubagentStart events in native Claude Code, `writeHookOutput` writes the context string directly to stdout.

When the event is `UserPromptSubmit`, this means the mode-change message (e.g., 'PONYTAIL MODE CHANGED — level: review') is written as hook output, which in Claude Code's hook system replaces the user's actual prompt with that text — the user's input is lost.

The `SubagentStart` case correctly wraps output in `{hookSpecificOutput:{...}}` to avoid this, but the `UserPromptSubmit` path does not.

nanaubusiness:
For native Claude Code, writing raw context to stdout on UserPromptSubmit replaces the user's prompt with the mode-change message. Using the JSON hookSpecificOutput form (already used for SubagentStart) properly supplements the prompt instead of replacing it.

Closes #363

### DIFF
--- a/.claude/instructions.md
+++ b/.claude/instructions.md
@@ -1,0 +1,3 @@
+---
+---
+<!-- This file is a fallback for CodeWhale. The canonical rules are in AGENTS.md. -->
--- a/after-install.md
+++ b/after-install.md
@@ -19,4 +19,4 @@ Commands:
 - `/ponytail-gain`
 - `/ponytail-help`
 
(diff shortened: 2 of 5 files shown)


DietrichGebert:
Closing: #363 has since been closed as resolved, so this PR is no longer needed. Thanks for the contribution — if you think the fix didn't fully land, comment and we'll re-open.

PR #382, opened 2026-06-27
fix: add 'rate limiting' to identifyTask for correctness gate

### LINKED ISSUE #372: benchmarks/correctness.js: identifyTask checks 'rate limit' but benchmark says 'rate limiting'
In `benchmarks/correctness.js` line 31:

```javascript
if (t.includes('rate limit') || t.includes('rate-limit')) return 'ratelimit';
```

But `benchmarks/promptfooconfig.yaml` line 41 defines the task as:
```
- vars: { task: "Add rate limiting to my FastAPI endpoint..." }
```

The word is 'rate limiting' (with ING), not 'rate limit' (space). The `identifyTask` check fails silently, so the rate-limiting correctness check is never run against the benchmark output.

nanaubusiness:
The benchmark prompt uses 'rate limiting' (with ING) but identifyTask only checked 'rate limit' and 'rate-limit'. The correctness gate was silently skipped for rate limit tasks.

Closes #372

### DIFF
--- a/benchmarks/correctness.js
+++ b/benchmarks/correctness.js
@@ -28,7 +28,7 @@ function identifyTask(task) {
   if (t.includes('debounce')) return 'debounce';
   if (t.includes('csv') && t.includes('sum')) return 'csv';
   if (t.includes('countdown') && t.includes('react')) return 'countdown';
-  if (t.includes('rate limit') || t.includes('rate-limit')) return 'ratelimit';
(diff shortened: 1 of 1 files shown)


DietrichGebert:
Thanks for the PR! Current main already classifies this correctly: identifyTask matches the substring 'rate limit', and since 'rate limiting'.includes('rate limit') is true the task is already picked up (benchmarks/correctness.js:31 also handles the 'rate-limit' form). So this change is redundant now. Closing alongside #372, but thanks for catching and following up on it.

PR #383, opened 2026-06-27
fix: only delete ponytail's own statusLine.command not the whole key

### LINKED ISSUE #374: scripts/uninstall.js: substring match deletes entire statusLine even for combined commands
In `scripts/uninstall.js` lines 28-34:

```javascript
const cmd = settings.statusLine && settings.statusLine.command;
if (typeof cmd === 'string' && cmd.includes('ponytail-statusline')) {
    delete settings.statusLine;
```

Uses substring match on the script name. If another plugin's statusline command is combined with ponytail's (e.g., `bash /caveman-statusline.sh && bash /ponytail-statusline.sh`), deleting the entire `statusLine` key destroys the other plugin's statusline. Fix: only delete ponytail's own entry, don't nuke the whole key.

nanaubusiness:
Previously uninstall deleted the entire statusLine key when it found 'ponytail-statusline' in the command string. Combined statuslines (e.g. caveman+ponytail) would lose other plugins' statuslines. Now only deletes the command field.

Closes #374

### DIFF
--- a/scripts/uninstall.js
+++ b/scripts/uninstall.js
@@ -31,7 +31,7 @@ try {
   // "ponytail-statusline" gets removed wholesale. Parse out only ponytail's part
   // if combined statuslines become common.
   if (typeof cmd === 'string' && cmd.includes('ponytail-statusline')) {
-    delete settings.statusLine;
+    delete settings.statusLine.command;
     fs.writeFileSync(settingsPath, JSON.stringify(settings, null, 2), 'utf8');
     console.log(`Removed ponytail statusLine entry from ${settingsPath}`);
   }


DietrichGebert:
Thanks @nanaubusiness. Digging into this one, delete settings.statusLine.command does not actually preserve the other plugin statusline: on a combined caveman && ponytail command it still drops the whole command string (caveman included) and leaves a {type:command} husk. I shipped #479 (merged) which only removes the statusLine when ponytail owns the whole command and otherwise leaves it untouched, with a regression test for the combined case. Credited you as co-author. Closing this in favor of #479.

PR #384, opened 2026-06-27
fix: strip UTF-8 BOM before JSON.parse in config

### LINKED ISSUE #375: hooks/ponytail-config.js: JSON.parse doesn't strip BOM — config silently fails on Windows
In `hooks/ponytail-config.js` line 86:

```javascript
const config = JSON.parse(fs.readFileSync(configPath, 'utf8'));
```

Unlike `ponytail-activate.js` (which strips UTF-8 BOM at line 49) and `check-versions.js` (which strips BOM at line 34), this function does not strip the BOM before JSON parsing. Config files saved with a BOM (common on Windows) cause silent parse failures, making user config settings be ignored with no warning.

nanaubusiness:
Unlike ponytail-activate.js and check-versions.js, this function did not strip BOM before parsing. Config files saved with BOM (common on Windows) silently failed to parse, ignoring user settings.

Closes #375

### DIFF
--- a/hooks/ponytail-config.js
+++ b/hooks/ponytail-config.js
@@ -83,7 +83,8 @@ function getDefaultMode() {
   // 2. Config file
   try {
     const configPath = getConfigPath();
-    const config = JSON.parse(fs.readFileSync(configPath, 'utf8'));
+    const raw = fs.readFileSync(configPath, 'utf8').replace(/^﻿/, '');
+    const config = JSON.parse(raw);
     if (config.defaultMode && VALID_MODES.includes(config.defaultMode.toLowerCase())) {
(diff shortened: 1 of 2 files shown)


DietrichGebert:
Thanks @nanaubusiness. I reconciled the #375 fix as #478 (merged) and credited you as co-author. Two changes from this PR: it uses the repo /^﻿/ escape rather than a literal BOM character in the source (an invisible source char is fragile and can be stripped by editors/linters), and it drops the scripts/uninstall.js hunk here since that duplicates your dedicated PR #383 (which belongs to #374 and stays open on its own). Closing this in favor of #478.

PR #393, opened 2026-06-27
fix: copilot-plugin test checks all 6 command files

nanaubusiness:
Previously only checked 4 of 6 command files (missing ponytail-gain.toml and ponytail-help.toml). Now validates all 6.

Closes #381

PR #394, opened 2026-06-27
fix: add error handling to setMode + check all 6 copilot commands

### LINKED ISSUE #378: hooks/ponytail-runtime.js: setMode has no try/catch — disk full crashes hook
In `hooks/ponytail-runtime.js` lines 15-18:

```javascript
function setMode(mode) {
  fs.mkdirSync(path.dirname(statePath), { recursive: true });
  fs.writeFileSync(statePath, mode);
}
```

No try/catch. If disk is full, permissions are wrong, or path is invalid, this throws and crashes the hook. Should gracefully degrade instead of propagating a crash.

### LINKED ISSUE #381: tests/copilot-plugin.test.js: only checks 4 of 6 commands for TOML file presence
In `tests/copilot-plugin.test.js` lines 11-16, only `ponytail.toml`, `ponytail-review.toml`, `ponytail-audit.toml`, and `ponytail-debt.toml` are checked. `ponytail-gain.toml` and `ponytail-help.toml` are not validated. If either file were deleted, the test would pass but the Copilot plugin would fail at runtime for those commands.

nanaubusiness:
Two fixes:
- `hooks/ponytail-runtime.js`: wrap setMode writeFileSync in try/catch so disk full/permission errors don't crash the hook (Closes #378)
- `tests/copilot-plugin.test.js`: check all 6 TOML files, not just 4 (Closes #381)

### DIFF
--- a/hooks/ponytail-runtime.js
+++ b/hooks/ponytail-runtime.js
@@ -13,8 +13,12 @@ if (isCopilot) stateDir = process.env.COPILOT_PLUGIN_DATA;
 const statePath = path.join(stateDir, STATE_FILE);
 
 function setMode(mode) {
-  fs.mkdirSync(path.dirname(statePath), { recursive: true });
-  fs.writeFileSync(statePath, mode);
+  try {
+    fs.mkdirSync(path.dirname(statePath), { recursive: true });
+    fs.writeFileSync(statePath, mode);
+  } catch (e) {
+    // Fail silently rather than crash the hook
+  }
 }
 
 function clearMode() {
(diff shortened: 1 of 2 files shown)


DietrichGebert:
Thanks @nanaubusiness, closing without merging. Two reasons:

1. The setMode try/catch is redundant. Both call sites already guard it: activate.js wraps setMode in try/catch ("flag is best-effort, do not block the hook"), and mode-tracker.js calls it inside finish()`s try/catch. Verified live on main by forcing the write to fail: both hooks exit 0, no crash. A third guard inside setMode adds nothing. Details on #378.
2. The copilot-plugin.test.js hunk (all 6 TOML files) already landed in main via #393, so that part is stale and would duplicate/conflict.

Appreciate the reports regardless.

PR #395, opened 2026-06-27
fix: add 'review' to RUNTIME_MODES so normalizeMode accepts it

### LINKED ISSUE #377: hooks/ponytail-config.js: 'review' in VALID_MODES but not RUNTIME_MODES — config accepts it but silently falls back to 'full'
In `hooks/ponytail-config.js`:

```javascript
const VALID_MODES = ['off', 'lite', 'full', 'ultra', 'review'];
const RUNTIME_MODES = ['off', 'lite', 'full', 'ultra'];
```

Setting `PONYTAIL_DEFAULT_MODE=review` passes validation, but `normalizeMode('review')` returns null (since 'review' is not in RUNTIME_MODES), causing fallback to 'full'. A user who configures 'review' as default would never know they're actually running in 'full' mode.

nanaubusiness:
Previously VALID_MODES included 'review' but RUNTIME_MODES did not. Setting PONYTAIL_DEFAULT_MODE=review passed config validation but normalizeMode('review') returned null, silently falling back to 'full'. Now 'review' is consistently accepted in both contexts.

Closes #377

### DIFF
--- a/hooks/ponytail-config.js
+++ b/hooks/ponytail-config.js
@@ -15,7 +15,7 @@ const os = require('os');
 
 const DEFAULT_MODE = 'full';
 const VALID_MODES = ['off', 'lite', 'full', 'ultra', 'review'];
-const RUNTIME_MODES = ['off', 'lite', 'full', 'ultra'];
+const RUNTIME_MODES = ['off', 'lite', 'full', 'ultra', 'review'];
 
 function normalizeMode(mode) {
   if (typeof mode !== 'string') return null;


DietrichGebert:
Closing without merging. The premise here is the issue: review is not an always-on runtime mode, it is a command-triggered task mode (run via /ponytail-review for a single over-engineering pass, then you return to your level). RUNTIME_MODES lists the always-on intensity levels a user selects (lite/full/ultra/off) and deliberately excludes review; VALID_MODES additionally includes review only so the command-set value persists and round-trips. Promoting review to RUNTIME_MODES would make normalizeMode treat it as an intensity level, which it is not — so this is not a missing entry to add.

(For the record, PONYTAIL_DEFAULT_MODE=review also does not fall back to full — it emits review instructions — but that is beside the main point.)

Thanks @nanaubusiness for looking regardless.

PR #396, opened 2026-06-27
fix: add main and exports to ponytail-mcp/package.json

### LINKED ISSUE #390: ponytail-mcp/package.json: no main or exports field
In `ponytail-mcp/package.json`, there is no `main` or `exports` field. The `index.js` is not reachable via npm module resolution (e.g., `require('@dietrichgebert/ponytail/ponytail-mcp')`). Users following the README MCP setup instructions work around this by pointing config directly at the file, but package-level imports would fail.

nanaubusiness:
Adds proper `main` and `exports` field so the MCP server package can be imported via npm module resolution.

Closes #390

### DIFF
--- a/ponytail-mcp/package.json
+++ b/ponytail-mcp/package.json
@@ -2,6 +2,10 @@
   "name": "ponytail-mcp",
   "version": "4.8.3",
   "description": "MCP server that serves Ponytail's lazy-senior-dev instructions as a prompt and a tool.",
+  "main": "./index.js",
+  "exports": {
+    ".": "./index.js"
+  },
   "private": true,
   "type": "module",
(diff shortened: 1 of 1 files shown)


DietrichGebert:
Thanks @nanaubusiness, closing without merging. ponytail-mcp is a private ESM MCP server that hosts spawn by path (README: { "command": "node", "args": ["ponytail-mcp/index.js"] }), not a package anything imports. main / exports are resolution hints for require/import, so there is nothing here they would fix, and the require("@dietrichgebert/ponytail/ponytail-mcp") pattern would not work regardless (ESM + private subdir). Details on #390. Appreciate the look.

PR #402, opened 2026-06-27
fix: use gpt-4o-mini instead of invalid gpt-5.4-mini

### LINKED ISSUE #399: benchmarks/robustness-audit.js: hardcodes invalid model gpt-5.4-mini as default

nanaubusiness:
Closes #399

### DIFF
--- a/benchmarks/robustness-audit.js
+++ b/benchmarks/robustness-audit.js
@@ -20,7 +20,7 @@ function python() {
 }
 
 const N = Number(process.env.AUDIT_N) || 20;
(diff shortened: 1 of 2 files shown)


DietrichGebert:
Closing without merging. gpt-5.4-mini is a valid, current OpenAI model (callable alias; snapshot gpt-5.4-mini-2026-03-17), their strongest mini per OpenAI docs: https://developers.openai.com/api/docs/models/gpt-5.4-mini

#399 is invalid, so this change downgrades a working default to an older model for no benefit. It also touches ponytail-mcp/index.js, which is unrelated to the stated fix. Closing in favor of leaving the benchmark default as-is.

PR #403, opened 2026-06-27
fix: use valid OpenAI model IDs in model-email.js

### LINKED ISSUE #400: benchmarks/model-email.js: hardcodes invalid OpenAI model IDs (gpt-4.1-mini, gpt-5.4-mini)
Line 7 hardcodes invalid model IDs:\n\n\n\nNeither  nor  are valid OpenAI model IDs. When ME_MODELS is not set, all API calls will fail with 400 errors. Should use valid model IDs (e.g. ) or require ME_MODELS to be set explicitly.\n\nNote: same bug as #388 which covers promptfooconfig.gpt.yaml, but this file was not covered.

nanaubusiness:
Replace invalid gpt-4.1-mini,gpt-5.4-mini with gpt-4o-mini,gpt-4o.

Closes #400


DietrichGebert:
Closing without merging. #400 was closed as not-a-bug: gpt-5.4-mini and gpt-4.1-mini are valid, current OpenAI models (verified against OpenAI docs). Swapping to gpt-4o-mini would downgrade a working default without fixing anything. Same reasoning as #402 and #510.

PR #416, opened 2026-06-27
fix: add commands/ to npm package files array

nanaubusiness:
## Bug

`package.json` `files` array does not include `commands/`. The `commands/` directory contains TOML command files (`ponytail.toml`, `ponytail-review.toml`, etc.) needed by Gemini CLI and Claude Code.

When a user installs `@dietrichgebert/ponytail` from npm, `commands/` is excluded from the installed package, breaking both hosts.

## Fix

Add `commands/` to the `files` array in `package.json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
   },
   "files": [
     "AGENTS.md",
+    "commands/",
     "hooks/",
     "skills/",
     ".opencode/",


DietrichGebert:
Thanks for the PR. From what I can see nothing installed from npm actually reads the root commands/ folder (opencode uses .opencode/command, pi registers commands in code, and Claude and Gemini install from git), so this would just ship unused files. Happy to revisit if I missed a consumer.

PR #421, opened 2026-06-27
fix: add commands/ to npm files array; align gemini-extension.test.js…

nanaubusiness:
… VERSIONED_MANIFESTS with check-versions.js

- package.json: commands/ was missing from files, breaking npm installs (Gemini CLI and Claude Code TOML commands not shipped)
- tests/gemini-extension.test.js: VERSIONED_MANIFESTS was missing .devin-plugin/plugin.json, package.json, and ponytail-mcp/package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
   },
   "files": [
     "AGENTS.md",
+    "commands/",
     "hooks/",
     "skills/",
     ".opencode/",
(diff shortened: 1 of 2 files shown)


DietrichGebert:
Thanks for this. Nothing installed from npm actually reads the root commands/ folder (opencode uses .opencode/command, pi registers commands in code, and Claude and Gemini install from git), so adding it to the files array would just ship unused files. This overlaps #416 which does the same thing, and I am closing both for that reason. Happy to revisit if I missed a consumer that needs it.

Issue #404, opened 2026-06-27
ponytail-help/SKILL.md: line 63 has wrong npm package name


nanaubusiness:
skills/ponytail-help/SKILL.md, Line 63

**Bug**: The instructions tell users to install Claude Code with:
```bash
npm install -g @anthropic-ai/claude-code@latest
```

**Problem**: The correct npm package name is `@anthropic-ai/code`, not `@anthropic-ai/claude-code`. The package `@anthropic-ai/claude-code` does not exist on npm.

**Fix**: Change `@anthropic-ai/claude-code@latest` to `@anthropic-ai/code@latest`

nanaubusiness:
FALSE POSITIVE. @anthropic-ai/claude-code IS the correct npm package — it exists on npm (version 2.1.195). @anthropic-ai/code does NOT exist. The skill instructions are correct.

PR #424, opened 2026-06-27
Fix wrong npm package name in update instructions

nanaubusiness:
Fixes the npm package name in the update instructions from the non-existent `@anthropic-ai/claude-code` to the correct `@anthropic-ai/code`.

See issue #404.


nanaubusiness:
FALSE POSITIVE. @anthropic-ai/claude-code IS the correct npm package (exists, version 2.1.195). @anthropic-ai/code does not exist. The skill instructions are correct. Reverting.

PR #431, opened 2026-06-27
fix: align ponytail-debt.md em-dash with TOML

nanaubusiness:
## Bug

The OpenCode command adapter `.opencode/command/ponytail-debt.md` used a comma () instead of an em dash (`—`) in the output format template, drifting from the canonical TOML at `commands/ponytail-debt.toml`.

**Before:** `<file>:<line>, <what was simplified>`
**After:** `<file>:<line> — <what was simplified>`

This was caught by a content drift check between the TOML and MD command adapters.


DietrichGebert:
Thanks for spotting the drift! It goes the other way though: the em dash was deliberately removed from the .md (repo convention), and the leftover one in the TOML is the actual bug. I'll fix the TOML side. Closing this one.

Issue #440, opened 2026-06-28
pi-extension: event.systemPrompt missing returns literal 'undefined' prepended to prompt


nanaubusiness:
In `pi-extension/index.js` lines 185-188:

```
return { systemPrompt: `${event.systemPrompt}\n\n${getPonytailInstructions(currentMode)}` };
```

If `event` is a truthy object without a `systemPrompt` key (e.g., `{}`), template literal stringification produces the literal string `"undefined"` prepended to every system prompt: `"undefined\n\nPONYTAIL MODE ACTIVE..."`. The model sees this garbage prefix on every turn.

DietrichGebert:
Fixed in #482. before_agent_start read event.systemPrompt with no guard, so a null/undefined event crashed with a TypeError (#439) and an event without a systemPrompt key stringified to the literal "undefined" and prepended it to every prompt (#440). Both are handled with one optional-chain guard that mirrors the event?. pattern the input handler already uses: it prepends the base prompt only when present, otherwise injects the ruleset alone. Regression test covers null/undefined event, missing systemPrompt, and the normal path (it fails on the old code with the exact TypeError). Thanks for the reports.

PR #444, opened 2026-06-29
fix: guard before_agent_start against missing event

### LINKED ISSUE #439: pi-extension: before_agent_start crashes if event is null/undefined

```
pi.on("before_agent_start", async (event) => {
  if (!currentMode || currentMode === "off") return;
});
```

`event.systemPrompt` is accessed without a null/undefined guard. If the pi runtime calls this handler with a falsy event (null/undefined), it throws `TypeError: Cannot read property 'systemPrompt' of null/undefined` and crashes the agent.

Note: the `input` handler at line 159 uses `event?.source` (optional chaining) but `before_agent_start` uses bare `event.systemPrompt` — inconsistent.

Aroool:
Fixes #439

This adds a null/undefined guard in the pi-extension `before_agent_start` handler so the extension does not crash if the Pi runtime invokes the hook without an event object.

The change intentionally only handles a missing event object. Handling a missing `event.systemPrompt` is left for #440.

Tested with:

```bash
cd pi-extension
npm test
```

Result:

```text
16 pass
0 fail
```

### DIFF
--- a/pi-extension/index.js
+++ b/pi-extension/index.js
@@ -183,7 +183,7 @@ export default function ponytailExtension(pi) {
   });
 
   pi.on("before_agent_start", async (event) => {
-    if (!currentMode || currentMode === "off") return;
+    if (!currentMode || currentMode === "off" || !event) return;
     return { systemPrompt: `${event.systemPrompt}\n\n${getPonytailInstructions(currentMode)}` };
   });
 }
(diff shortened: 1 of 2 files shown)


DietrichGebert:
Thanks @Aroool. Your guard was the right instinct, but it only covered the null/undefined event (#439) and left #440 open, and the !event check does not catch the {} case (a truthy event with no systemPrompt still stringifies to "undefined"). I reconciled a single optional-chain guard that closes both #439 and #440 in #482 (merged) and credited you as co-author, with a regression test that fails on the old code with the exact TypeError. Closing this in favor of #482. Appreciate the fix.

PR #446, opened 2026-06-29
feat: add kimi-code support

VinkyDev:


DietrichGebert:
Thanks for adding Kimi. Two things: Kimi Code loads skills as SKILL.md directories (via `--skills-dir` or `extra_skill_dirs`) and reads AGENTS.md, and its plugin system is MCP config, so it does not actually read a `.kimi-plugin/plugin.json`, this file would just sit unused. And the repo already ships AGENTS.md plus `skills/`, so Kimi works today with no new files. The right addition is a short docs note, not a manifest, so I am going to close this.

For anyone landing here, using ponytail with Kimi Code CLI today:

- **Rules:** Kimi reads `AGENTS.md` from the project root, so this repo's `AGENTS.md` gives you the always-on ruleset.
- **Skills:** point Kimi at the bundled skills with `kimi --skills-dir skills/`, or add `skills/` to `extra_skill_dirs` in your config to make it permanent (this registers `/ponytail`, `/ponytail-review`, and the rest).

Thanks for the push, it made me go check how Kimi actually loads.

PR #452, opened 2026-06-29
feat(debt): flag missing upgrade triggers and scan defer: comments

kevglynn:
## Summary

- Scans both `ponytail:` and `defer:` comment prefixes (dual-prefix support)
- Explicitly flags entries missing upgrade triggers with `⚠ no-trigger` tag
- Adds well-formed vs missing-trigger examples
- Reports `<N> markers, <M> with no trigger` in summary

## Motivation

Projects integrating ponytail alongside other frameworks may use a tool-neutral `defer:` prefix instead of `ponytail:`. The debt harvester should find both. Additionally, the no-trigger detection was implicit — making it explicit with examples helps agents identify rot risk.

## Changes

- `skills/ponytail-debt/SKILL.md`: dual-prefix scan, explicit no-trigger flagging with examples
- `.openclaw/skills/ponytail-debt/SKILL.md`: regenerated from canonical source
- `commands/ponytail-debt.toml`: updated Claude command prompt
- `.opencode/command/ponytail-debt.md`: updated OpenCode command prompt

## Test plan

- [x] `check-rule-copies.js` passes
- [x] `npm test` passes (1 pre-existing unrelated failure in csv probe)
- [ ] Manual: create test files with both `ponytail:` and `defer:` comments (some with triggers, some without) — verify harvester catches both and flags missing triggers


kevglynn:
**Updated:** Stripped out the `defer:` dual-prefix scanning — that was our internal convention leaking into the PR. The force-pushed commit now focuses only on the no-trigger detection enhancement:

- Explicit `⚠ no-trigger` tagging for `ponytail:` comments missing upgrade paths or ceilings
- Well-formed vs missing-trigger examples in the skill
- Updated command files and OpenClaw copies

No changes to the existing `ponytail:` convention itself. All tests pass (68/69, 1 pre-existing csv/pandas).

kevglynn:
Checking in and rebased onto latest main to keep this current. Let me know if you'd like any changes to the no-trigger detection approach, or if this doesn't fit where you want ponytail-debt to go.

PR #453, opened 2026-06-29
fix: prevent Windows session freeze from stdin EOF deadlock (#443)

### LINKED ISSUE #443: Bug: Claude Code freezes indefinitely on Windows after installing Ponytail
# 🐛 Bug: Claude Code freezes indefinitely on Windows after installing Ponytail

## 💻 Environment

* **OS:** Windows 10 (PowerShell 5.1)
* **Node.js:** v18+
* **Claude Code:** Latest (VSCode extension)
* **Ponytail:** Installed via `npm install -g @dietrichgebert/ponytail`

---

## 🛑 Symptoms

After installing Ponytail, Claude Code **freezes completely** when submitting any prompt. The terminal becomes unresponsive — no spinner, no output, no error. Only `Ctrl+C` recovers the session.

A secondary issue occurs on first launch: 
> `Failed to set model: Unexpected token '', "{ "p"... is not valid JSON`

This is caused by a UTF-8 BOM prepended by PowerShell when writing `settings.json`.

---

## 🔍 Root Cause Analysis

Two distinct issues were identified during investigation:

### 1. Stdin EOF deadlock (Critical — causes the freeze)
`ponytail-mode-tracker.js` (the `UserPromptSubmit` hook) reads stdin using the following pattern:

```js
process.stdin.on('end', () => {
  // parse JSON, process command, exit
});

### PR DESCRIPTION
**Pull request by Hasnan42**
## Problem

Closes #443. On Windows, Claude Code freezes indefinitely after installing ponytail — submit any prompt and the session hangs with no spinner, no output, no error; only `Ctrl+C` recovers it.

## Root cause

`hooks/ponytail-mode-tracker.js` (the `UserPromptSubmit` hook) only ever completed inside `process.stdin.on('end', …)`:

```js
process.stdin.on('end', () => { /* parse, set mode, exit */ });
```

On Windows the hook is launched via the PowerShell wrapper in `claude-codex-hooks.json`:

```
if (Get-Command node -ErrorAction SilentlyContinue) { node "$env:CLAUDE_PLUGIN_ROOT\hooks\ponytail-mode-tracker.js" }
```

A native command invoked inside a PowerShell `if {}` block does not reliably receive the parent's piped stdin, so the prompt JSON never reaches `node` and the `'end'` event never fires. The process stays alive waiting on stdin forever, and Claude Code blocks on the hook — freezing the session.

## Fix

Make the hook non-blocking, matching the best-effort / never-block contract the other lifecycle hooks already follow:

- Extract the handler into `finish()`, guarded by a `done` flag so it runs at most once.
- Add a `process.stdin.on('error', …)` handler.
- Add a **1s `unref()`'d safety timeout** that runs `finish()` (recovering the mode even if data arrived without EOF) and exits cleanly.

`unref()` means the timer never holds the event loop open on the normal path — `'end'` fires first there, so there is **zero added latency** for the common case. The fallback only matters when EOF never comes.

## Testing

- `node --test tests/*.test.js`: same results as `main` except **+1 passing test** — no new failures. (Pre-existing failures in `hermes-plugin`/`correctness` only occur where Python/pandas isn't installed locally and are unrelated.)
- Added a regression test in `tests/hooks-windows.test.js` that spawns the hook with a stdin pipe that is never written to or closed, and asserts the process self-exits (cleanly, code 0) instead of hanging. Without this fix the test hits its 3s guard and fails.

Manually verified: spawning the patched hook with an open, empty stdin pipe exits in ~1s (code 0) rather than hanging indefinitely.

### DIFF
--- a/hooks/ponytail-mode-tracker.js
+++ b/hooks/ponytail-mode-tracker.js
@@ -6,8 +6,11 @@ const { getDefaultMode, isDeactivationCommand } = require('./ponytail-config');
 const { clearMode, setMode, writeHookOutput } = require('./ponytail-runtime');
 
 let input = '';
-process.stdin.on('data', chunk => { input += chunk; });
-process.stdin.on('end', () => {
+let done = false;
+
+function finish() {
+  if (done) return;
+  done = true;
   try {
     // Strip UTF-8 BOM some shells prepend when piping (breaks JSON.parse)
     const data = JSON.parse(input.replace(/^\uFEFF/, ''));
@@ -52,4 +55,17 @@ process.stdin.on('end', () => {
   } catch (e) {
     // Silent fail
   }
-});
+}
+
+process.stdin.on('data', chunk => { input += chunk; });
+process.stdin.on('end', finish);
+
+// Never hang the session. On Windows, Claude Code runs this hook through a
+// PowerShell `if {}` wrapper that can swallow the piped prompt JSON, so stdin
+// 'end' never fires and the hook blocks forever — freezing the session (#443).
+// On error, or after a short fallback, process whatever arrived (recovering the
+// mode if data came without EOF) and exit. unref() keeps the timer from adding
+// latency to the normal path, where 'end' fires first. Mirrors the best-effort,
+// never-block contract the other lifecycle hooks already follow.
+process.stdin.on('error', () => { finish(); process.exit(0); });
+setTimeout(() => { finish(); process.exit(0); }, 1000).unref();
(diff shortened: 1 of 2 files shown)


DietrichGebert:
Thanks @Hasnan42, this is a clean, well-reasoned fix. It branched before #227 (stdin error handler) and #474 (exec node hooks) landed, so it conflicts on both changed files. I reconciled the identical fix onto current main in #477 and credited you as co-author. Closing this in favor of #477, which is merged. Appreciate the thorough writeup and the regression test.

PR #459, opened 2026-06-29
fix(pi): avoid undefined system prompt prefix

### LINKED ISSUE #440: pi-extension: event.systemPrompt missing returns literal 'undefined' prepended to prompt

terminalchai:
Treats a missing Pi `event.systemPrompt` as an empty base prompt before appending Ponytail instructions, so a truthy event object like `{}` no longer injects a literal `undefined` prefix into the model's system prompt.

Closes #440.

Verified with `npm test --prefix pi-extension` and full `npm test` after prepending the local Python 3.11 install to PATH so the Hermes tests can resolve `python3`.

### DIFF
--- a/pi-extension/index.js
+++ b/pi-extension/index.js
@@ -184,6 +184,8 @@ export default function ponytailExtension(pi) {
 
   pi.on("before_agent_start", async (event) => {
     if (!currentMode || currentMode === "off") return;
-    return { systemPrompt: `${event.systemPrompt}\n\n${getPonytailInstructions(currentMode)}` };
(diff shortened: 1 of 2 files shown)


DietrichGebert:
Closing: #440 (and #439) are already fixed in #482 (merged), which avoids the undefined systemPrompt prefix and the null-event crash. Thanks!

PR #462, opened 2026-06-30
fix(pi): guard against null/undefined event in before_agent_start (closes #439)

### LINKED ISSUE #439: pi-extension: before_agent_start crashes if event is null/undefined

Tanmay9223:
🎯 **What:** Adds `event?.systemPrompt ?? ""` optional chaining in the `before_agent_start` handler in `pi-extension/index.js`, replacing the bare `event.systemPrompt` property access.

💡 **Why:** If the pi runtime calls `before_agent_start` with a falsy event (null or undefined), the handler immediately throws `TypeError: Cannot read properties of null (reading 'systemPrompt')`, crashing the agent before every session turn. The adjacent `input` handler already uses `event?.source` and `event?.text` defensively — `before_agent_start` was the only handler missing this guard (issue #439).

✅ **Verification:** Added three targeted regression tests to `pi-extension/test/extension.test.js` covering: null event, undefined event, and a truthy event object without a `systemPrompt` key. All three were failing before the fix and pass after. Full pi-extension suite: 18/18 pass. Full repo suite: 68/69 pass — the single remaining failure is the pre-existing `csv: correct pandas one-liner passes` test, which requires a `pandas` install and was already failing on `main` before this change.

✨ **Result:** The `before_agent_start` handler is now robust to any event shape the pi runtime may pass, matching the defensive optional-chaining style used throughout the rest of the extension.

### DIFF
--- a/pi-extension/index.js
+++ b/pi-extension/index.js
@@ -184,6 +184,7 @@ export default function ponytailExtension(pi) {
 
   pi.on("before_agent_start", async (event) => {
     if (!currentMode || currentMode === "off") return;
-    return { systemPrompt: `${event.systemPrompt}\n\n${getPonytailInstructions(currentMode)}` };
+    return { systemPrompt: `${event?.systemPrompt ?? ""}\n\n${getPonytailInstructions(currentMode)}` };
   });
+
 }
--- a/pi-extension/test/extension.test.js
+++ b/pi-extension/test/extension.test.js
@@ -165,3 +165,39 @@ test("status bar stays silent when ui lacks a theme", async () => withTempConfig
 
   assert.deepEqual(calls, []);
 }));
+
+// Regression: before_agent_start must not throw when event is null/undefined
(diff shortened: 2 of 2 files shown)


Tanmay9223:
Closing this PR as a duplicate. PR #444 by @Aroool already addresses the same issue and was opened before this one. My PR search failed to catch it because it used 'Fixes #439' rather than 'Closes #439'. Apologies for the noise.

PR #297, opened 2026-06-24
docs: add Chinese README translations

kylekkkk61:
## Summary

- Add native Simplified Chinese and Traditional Chinese README translations.
- Link both translations from the English README language selector.
- Preserve the English README's commands, URLs, structure, and benchmark claims.

## Validation

- Verified heading levels, URLs, and non-translated command blocks against `README.md`.
- Ran `git diff --check`.

No runtime, package, version, or rule changes.

PR #467, opened 2026-06-30
docs: add Simpified-Chinese README translation

DarkModest:
Co-Author: DeepSeek-v4-flash

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@
 </p>
 
 <p align="center">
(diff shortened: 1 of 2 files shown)


DietrichGebert:
Thanks so much for the Chinese README. There are a few of these open, so I am keeping one (#297, which covers both simplified and traditional) to avoid duplicates. Appreciate the effort.

PR #471, opened 2026-07-01
fix(package): ship uninstall cleanup script

Anmolnoor:
## What

Ships `scripts/uninstall.js` in the npm package.

## Why

This complements the fix from @isaukywhite in #469 for #434. That PR makes `scripts/uninstall.js` safer when `settings.json` is malformed, but the npm package currently does not ship `scripts/uninstall.js`, even though the README tells users to run it.

## How

Add `scripts/uninstall.js` to `package.json`'s `files` list and add a package regression test using `npm pack --dry-run --json`.

## Test

`node --test tests/package.test.js tests/uninstall.test.js`

Related: #469 by @isaukywhite
Refs #434

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "skills/",
     ".opencode/",
     "pi-extension/",
+    "scripts/uninstall.js",
     "assets/",
     "LICENSE"
   ],
--- a/tests/package.test.js
+++ b/tests/package.test.js
@@ -1,0 +1,32 @@
+#!/usr/bin/env node
+
+const test = require('node:test');
+const assert = require('node:assert/strict');
+const fs = require('fs');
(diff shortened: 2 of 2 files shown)


DietrichGebert:
Thanks @Anmolnoor, this was a real gap: the README tells users to run scripts/uninstall.js but the package did not ship it. I landed the fix in #538 (merged) and credited you as co-author. One change: I replaced the npm pack subprocess test with a plain files-array assertion. The pack test passed in CI but broke when run standalone on Windows (spawnSync on npm.cmd without shell:true throws EINVAL on modern Node), which your own test command would hit. The assertion guards the same regression on every platform with no subprocess. Closing this in favor of #538. Appreciate the catch.

PR #472, opened 2026-07-01
Add Simplified Chinese README (README.zh.md)

wuisabel-gif:
Full mirror translation of README.md following the es/ko convention: all structure, sections, install commands, tables and links preserved; only prose and image alt text translated. Adds a 中文 link to the language selector in README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@
 </p>
 
 <p align="center">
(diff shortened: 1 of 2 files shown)


DietrichGebert:
Thank you for the translation work. There are a few Chinese README PRs open right now, so I want to keep just one to avoid maintaining several copies. I am going with #297 since it adds both simplified and traditional. Really appreciate you helping make ponytail more readable.

PR #473, opened 2026-07-01
fix(opencode): add named exports for opencode loader compatibility

batuhanysv:
## Problem

opencode's plugin loader imports a plugin module and expects a **NAMED** plugin export — every documented example at https://opencode.ai/docs/plugins/ uses the shape:

```js
export const MyPlugin = async (ctx) => ({ ... })
```

`.opencode/plugins/ponytail.mjs` currently only provides `export default async (...) => {...}`. As a result, the OpenCode loader fails to wire the plugin's lifecycle hooks and logs on every startup:

```
failed to load plugin  path=@dietrichgebert/ponytail  error="path must be a string or a file descriptor"
```

(The slash commands still register, but the ruleset injection + `/ponytail <mode>` persistence — i.e. the `config`, `experimental.chat.system.transform`, and `command.execute.before` hooks — never fire.)

## Fix

Expose the existing plugin function as named exports in addition to the default export. Fully additive, non-breaking:

```diff
+const ponytailPlugin = async ({ client } = {}) => { ... }
+
+export { ponytailPlugin as ponytail };
+export { ponytailPlugin as Ponytail };
+export { ponytailPlugin as PonytailPlugin };
+export default ponytailPlugin;
-export default async ({ client } = {}) => { ... }
```

Three named aliases are provided because opencode's docs do not specify an exact expected export name; this covers the most idiomatic options (lowercased package name, PascalCase, and the conventional `PonytailPlugin` form used in examples). Hosts that import the default export (Claude Code, pi, Codex) keep working unchanged.

## Verification

Tested on opencode v1.15.7 with `{ "plugin": ["@dietrichgebert/ponytail"] }` in `opencode.jsonc`:

- **Before**: every startup logs `failed to load plugin  path=@dietrichgebert/ponytail  error="path must be a string or a file descriptor"` and the ponytail ruleset is never injected into the system prompt.
- **After**: the plugin loads cleanly, `/ponytail`, `/ponytail-help`, `/ponytail-audit`, `/ponytail-gain`, `/ponytail-debt`, `/ponytail-review` all register, and the ponytail ruleset is injected at the active level every turn.

## Scope

One-file, behavior-identical to current default export — no logic changes, no new deps, no breaking changes for any existing host.

### DIFF
--- a/.opencode/plugins/ponytail.mjs
+++ b/.opencode/plugins/ponytail.mjs
@@ -51,7 +51,7 @@ export function parseCommandFile(filePath) {
   return { description, template: match[2].trim() };
 }
 
-export default async ({ client } = {}) => {
+const ponytailPlugin = async ({ client } = {}) => {
   const log = (level, message) => {
     try { client && client.app && client.app.log({ body: { service: 'ponytail', level, message } }); } catch (e) {}
   };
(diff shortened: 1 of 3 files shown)


DietrichGebert:
Thanks for digging into the OpenCode loader, it is genuinely confusing behavior. I checked how it actually loads plugins though, and the real issue is the opposite of named exports: OpenCode's loader iterates over every export from the plugin module and calls each one as a plugin function (opencode issue #13543, "External plugins fail to load: TypeError: fn is not a function"). Named vs default is not the problem, it accepts either for the plugin function. The crash comes from the module also exporting the parseCommandFile helper, which the loader tries to call as a plugin. This PR keeps that export and adds three aliases plus an index.mjs indirection, so it does not fix the crash. #301 fixes it the right way by moving parseCommandFile into a .cjs sibling so the plugin module exports only the plugin function. Going with that one, so closing this. Really appreciate you chasing it.

PR #485, opened 2026-07-02
fix: handle CRLF line endings in loc.js fence regex (#339)

### LINKED ISSUE #339: benchmarks: loc.js misses CRLF code fences and counts the whole response

guyoron1:
Heyy, noticed `loc.js` was the odd one out — `correctness.js` and `robustness-audit.js` both handle CRLF with `\r?\n` but `loc.js` still used bare `\n`, so fenced blocks from Windows-origin responses just silently didn't match. Quick regex fix to bring it in line with the rest.

Also spotted that `codeOf()` in `behavior.js` is defined but never actually called by any of the check functions — removed the dead code while I was in there.

## Summary
- Fixes `loc.js` fence-matching regex to handle `\r\n` (CRLF) line endings using `\r?\n`
- Removes dead `codeOf()` function from `behavior.js` (defined but never called by any check function)
- Adds CRLF test case to `loc.test.js`

## Test plan
- [x] `node benchmarks/loc.test.js` — 6/6 pass including new CRLF test
- [x] `node tests/behavior.test.js` — 8/8 pass after dead code removal

Fixes #339

### DIFF
--- a/benchmarks/behavior.js
+++ b/benchmarks/behavior.js
@@ -10,10 +10,6 @@
 //
 // Metric: `behavior` (1 = behavior present, 0 = absent).
 
-function codeOf(text) {
-  return [...String(text || '').matchAll(/```[\w-]*\n([\s\S]*?)```/g)].map((m) => m[1]).join('\n');
-}
-
 function proseOf(text) {
   return String(text || '').replace(/```[\s\S]*?```/g, ' ').replace(/\s+/g, ' ').trim();
 }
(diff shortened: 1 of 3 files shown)


DietrichGebert:
Closing: this is byte-identical to #518, which was merged for #339 (and removes the dead codeOf, also resolving #533). Thanks for the fix.

PR #486, opened 2026-07-02
fix: detect compression-plugin conflicts and add deconfliction (#332)

### LINKED ISSUE #332: [Bug]: ponytail does not detect caveman conflict — both inject contradictory instructions
## Bug Description

When ponytail and caveman are both active, they both inject `additionalContext` on `UserPromptSubmit`. The model receives two contradictory style instruction sets every turn:

- Caveman: `"CAVEMAN MODE ACTIVE (full). Drop articles/filler/pleasantries/hedging. Fragments OK."`
- Ponytail: `"PONYTAIL MODE ACTIVE (full). The ladder enforced. Stdlib and native first."`

Previous issues (#18, #68, #193) asked "can they coexist?" and were closed with "yes, they stack". But the **actual bug** is that both plugins inject per-turn style reminders simultaneously, causing the model to receive contradictory instructions about output style. This results in unpredictable, inconsistent outputs.

## What caveman already does

Caveman PR [#577](https://github.com/JuliusBrussee/caveman/pull/577) added conflict detection: when ponytail (or other compression plugins) is detected in `settings.json`, caveman reduces its per-turn injection to a one-liner instead of the full ruleset. But ponytail does NOT reciprocate — it always injects its full ruleset regardless of whether caveman is active.

## Proposed Fix

In `ponytail-mode-tracker.js`, before emitting the per-turn `additionalContext`, check `settings.json` for known compression plugins (caveman, etc.). If detected, reduce the per-turn injection to a minimal one-liner:

```
PONYTAIL MODE ACTIVE (full) — reduced injection (another compression plugin detected). Ponytail rules still apply per SessionStart injection.
```

The SessionStart full ruleset is unchanged — ponytail rules still apply, just the per-turn reinforcement is reduced to avoid contradictory instructions.

This mirrors the approach caveman already uses in its `detectCompressionConflict()` function.

## Testing

With caveman installed in settings.json:
- **Before**: model receives both `"CAVEMAN MODE ACTIVE (full). Drop articles..."` AND `"PONYTAIL MODE ACTIVE (full). The ladder enforced..."` every turn → contradictory instructions
- **After**: model receives `"CAVEMAN MODE ACTIVE (full) — reduced injection"` + ponytail full injection = no contradiction

guyoron1:
Hey! Saw that caveman already added ponytail detection on their side (PR #577), so figured ponytail should reciprocate. Right now if you have both installed, the model gets two full instruction sets with overlapping (and sometimes contradictory) style rules every session — not great.

This adds a `detectCompressionPlugins()` check that scans `settings.json` for known peers. When caveman is present, ponytail appends a short deconfliction paragraph establishing clear lanes: ponytail owns code structure decisions (the ladder, YAGNI, stdlib-first), peer plugin owns output style. No more tug-of-war.

## Changes
- `hooks/ponytail-config.js` — new `detectCompressionPlugins()` function + export
- `hooks/ponytail-instructions.js` — `getCompressionDeconfliction()` appends boundary paragraph when peers detected
- `tests/hooks.test.js` — tests detection with/without caveman in mock settings.json

## Test plan
- [x] `node tests/hooks.test.js` passes with new caveman detection tests
- [x] Verifies deconfliction paragraph appears when caveman detected
- [x] Verifies no deconfliction paragraph without caveman
- [x] Pi-extension tests (8/8) still pass

Fixes #332

### DIFF
--- a/hooks/ponytail-config.js
+++ b/hooks/ponytail-config.js
@@ -95,6 +95,24 @@ function getDefaultMode() {
   return DEFAULT_MODE;
 }
 
+const KNOWN_COMPRESSION_PLUGINS = ['caveman'];
+
+function detectCompressionPlugins() {
+  try {
+    const settingsPath = path.join(getClaudeDir(), 'settings.json');
+    const raw = fs.readFileSync(settingsPath, 'utf8').replace(/^﻿/, '');
+    const settings = JSON.parse(raw);
+    const found = [];
+    const blob = JSON.stringify(settings).toLowerCase();
+    for (const name of KNOWN_COMPRESSION_PLUGINS) {
+      if (blob.includes(name)) found.push(name);
+    }
+    return found;
+  } catch (e) {
+    return [];
+  }
+}
+
 function writeDefaultMode(mode) {
   const normalized = normalizeConfigMode(mode);
   if (!normalized) return null;
@@ -109,6 +127,7 @@ module.exports = {
   DEFAULT_MODE,
   VALID_MODES,
   RUNTIME_MODES,
+  detectCompressionPlugins,
   getDefaultMode,
   getConfigDir,
   getConfigPath,
--- a/hooks/ponytail-instructions.js
+++ b/hooks/ponytail-instructions.js
@@ -3,7 +3,7 @@
 
 const fs = require('fs');
 const path = require('path');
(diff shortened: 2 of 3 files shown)


DietrichGebert:
Thanks for putting this together. There is a sibling PR (#517) doing the same compression conflict detection and it merges cleanly right now, so we are landing that one instead. Thanks for digging into #332.

PR #487, opened 2026-07-02
fix: add CJK language paragraph to ponytail instructions (#333)

### LINKED ISSUE #333: [Bug]: No CJK handling — English style rules conflict with Chinese/mixed-language conversations
## Bug Description

Ponytail's style rules are designed for English-only conversations. When used in Chinese/Japanese/Korean (CJK) contexts, several issues arise:

1. **"Drop articles" instruction is misleading in CJK context**: Chinese has no article concept. When ponytail tells the model to drop articles, the model may incorrectly strip English articles embedded in Chinese text (e.g. "a LLM 模型" → "LLM 模型"), losing semantic specificity.

2. **"Short synonyms" rule assumes English**: The instruction "big not extensive, fix not implement a solution for" only works in English. In CJK, shorter synonyms don't exist in the same way — Chinese characters are already compact.

3. **Output style "fragments OK" is language-agnostic but the examples are English-only**: CJK users get English examples that don't map to their language patterns.

## Proposed Fix

Add CJK detection (similar to caveman PR [#576](https://github.com/JuliusBrussee/caveman/pull/576)):

1. Detect CJK characters in user input/session context
2. When CJK is detected, modify the per-turn injection to:
   - Skip article-dropping instructions (Chinese has no articles)
   - Keep code-focused rules (stdlib first, YAGNI, shortest diff) — these are language-agnostic
   - Add CJK-specific brevity guidance (e.g. "简短回答，不用敬语")

This is the same approach caveman uses: CJK detection → skip English-specific rules → keep language-agnostic rules.

## Related

- Caveman PR #576: Same issue fixed for caveman's compression engine
- #18, #68: Previous questions about i18n, closed as "works as designed"

guyoron1:
Heyy — was using ponytail in a mixed Chinese/English codebase and realized the "drop articles" and "short synonyms" rules don't really make sense for CJK. Chinese doesn't have articles, and CJK characters are already compact so "use shorter synonyms" is a no-op at best, confusing at worst.

Rather than adding runtime language detection in the hooks (which don't see user messages anyway), this takes the instruction-level approach similar to caveman PR #576 — just tells the model directly to skip the English-specific brevity rules when responding in CJK. The structural stuff (ladder, YAGNI, stdlib-first) is language-agnostic and still applies.

## Summary
- Adds a CJK languages section to ponytail instructions
- Skips: article dropping, synonym shortening (English-specific)
- Keeps: all structural rules (ladder, YAGNI, stdlib-first, shortest diff)
- Code comments/identifiers follow whichever language the codebase already uses

## Test plan
- [x] `node tests/hooks.test.js` passes with new CJK assertions
- [x] Pi-extension tests (8/8) still pass
- [x] Verified CJK paragraph present in both SKILL.md-based and fallback instruction paths

Fixes #333

### DIFF
--- a/hooks/ponytail-instructions.js
+++ b/hooks/ponytail-instructions.js
@@ -70,6 +70,13 @@ function getFallbackInstructions(mode) {
     'Ponytail governs what you build, not how you talk. "stop ponytail" or "normal mode": revert. Level persists until changed or session end.';
 }
 
+const CJK_PARAGRAPH =
+  '\n\n## CJK languages\n\n' +
+  'When the conversation is in Chinese, Japanese, or Korean: ' +
+  'skip English-specific brevity rules (article dropping, synonym shortening — CJK has no articles and is already compact). ' +
+  'The ladder, YAGNI, stdlib-first, shortest diff, and all structural rules still apply. ' +
+  'Keep code comments and identifiers in whichever language the codebase already uses.';
+
 function getPonytailInstructions(mode) {
   const configuredMode = normalizePersistedMode(mode) || DEFAULT_MODE;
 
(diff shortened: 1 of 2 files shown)


DietrichGebert:
Thanks a lot for taking on #333. Another PR (#516) makes the same CJK change and is already in a clean mergeable state, so we are going with that one. Really appreciate the effort.

PR #488, opened 2026-07-02

### LINKED ISSUE #329: pi ponytail on and off issue

guyoron1:
Was playing around with different modes and noticed that /ponytail off doesn't stick across sessions — next time Claude starts, it's right back to full. Turns out the mode tracker cleared the session flag but never persisted the choice to config.json, so SessionStart re-activated every time.

## Summary
- Import and call `writeDefaultMode()` in `ponytail-mode-tracker.js` when mode changes
- All runtime mode switches (lite, full, ultra, off) are now persisted to `~/.config/ponytail/config.json`
- Review mode is intentionally excluded from persistence (it's a transient workflow)
- Deactivation via "stop ponytail" / "normal mode" also persists "off"

## Test plan
- [x] Added end-to-end persistence test: `/ponytail off` → verify config.json → SessionStart respects it
- [x] Added re-enable test: `/ponytail lite` → verify config.json updated
- [x] All existing hook tests still pass (`node tests/hooks.test.js`)

Fixes #329

### DIFF
--- a/hooks/ponytail-mode-tracker.js
+++ b/hooks/ponytail-mode-tracker.js
@@ -2,7 +2,7 @@
 // ponytail — UserPromptSubmit hook to track which ponytail mode is active
 // Inspects user input for /ponytail commands and writes mode to flag file
 
-const { getDefaultMode, isDeactivationCommand } = require('./ponytail-config');
+const { getDefaultMode, isDeactivationCommand, writeDefaultMode } = require('./ponytail-config');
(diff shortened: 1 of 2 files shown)


DietrichGebert:
Closing in favor of #541 (merged). Same reasoning as on #515: persisting every switch breaks the documented session-scoped design, so #541 adds an explicit /ponytail default <mode> command instead. Your test coverage was good though. Thanks for the fix.

PR #489, opened 2026-07-02
fix: writeDefaultMode merges config instead of overwriting

### LINKED ISSUE #490: [Bug]: writeDefaultMode overwrites entire config.json instead of merging
## Description

`writeDefaultMode()` in `hooks/ponytail-config.js` creates a fresh `{ defaultMode: "..." }` object and writes it as the entire file contents. If `config.json` already had other fields (either from future features or user customizations), they get silently dropped.

## Steps to reproduce

1. Create `~/.config/ponytail/config.json` with extra fields:
   ```json
   { "defaultMode": "full", "customSetting": 42 }
   ```
2. Run `/ponytail ultra` (or any mode switch that calls `writeDefaultMode`)
3. Read config.json — `customSetting` is gone:
   ```json
   { "defaultMode": "ultra" }
   ```

## Expected behavior

`writeDefaultMode` should read-merge-write: load the existing config, update only `defaultMode`, and write it back preserving other fields.

## Actual behavior

The entire file is overwritten with only `{ "defaultMode": "..." }`.

guyoron1:
Was looking at the config persistence path and realized writeDefaultMode() creates a fresh `{ defaultMode: ... }` object every time it writes — if config.json had any other fields, they'd be silently dropped.

## Summary
- Changed `writeDefaultMode()` to read-merge-write: loads existing config.json, updates `defaultMode`, writes back
- Handles missing/malformed config gracefully (falls back to empty object)
- Strips UTF-8 BOM before parsing, matching the pattern used elsewhere in the codebase

## Test plan
- [x] Added test: seed config.json with `{ defaultMode: "full", customSetting: 42 }`, write "ultra", verify both fields survive
- [x] All existing hook tests pass (`node tests/hooks.test.js`)
- [x] Pi-extension tests pass (`node --experimental-vm-modules pi-extension/test/extension.test.js`)

Fixes #490

### DIFF
--- a/hooks/ponytail-config.js
+++ b/hooks/ponytail-config.js
@@ -102,7 +102,13 @@ function writeDefaultMode(mode) {
 
   const configPath = getConfigPath();
   fs.mkdirSync(path.dirname(configPath), { recursive: true });
-  fs.writeFileSync(configPath, JSON.stringify({ defaultMode: normalized }, null, 2), 'utf8');
+  let config = {};
+  try {
+    config = JSON.parse(fs.readFileSync(configPath, 'utf8').replace(/^﻿/, ''));
+    if (!config || typeof config !== 'object' || Array.isArray(config)) config = {};
+  } catch (_) {}
+  config.defaultMode = normalized;
+  fs.writeFileSync(configPath, JSON.stringify(config, null, 2), 'utf8');
   return normalized;
 }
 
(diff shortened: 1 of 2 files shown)


DietrichGebert:
Closing in favor of #514 (merged), which is the same read-merge-write fix for #490. I took #514 because it uses the /^﻿/ escape rather than a literal BOM character in the source (an invisible char is fragile and can be stripped by editors), and it does not add an unused import. Thanks for the fix.

PR #491, opened 2026-07-02
fix: tighten simplification marker rule

### LINKED ISSUE #120: Comment and commit behavior.
Does it always have to add "ponytail:" or "//ponytail:" in each code change and commit name?

sanmaxdev:
## Summary
- Tightens the simplification marker rule so normal or trivial code gets no `ponytail:` comment.
- Syncs the updated wording across the copied rule files, runtime instructions, and generated OpenClaw skill.
- Updates the Pi helper regression assertion for the new boundary text.

## Testing
- [x] `node scripts/check-rule-copies.js`
- [x] `node scripts/check-versions.js`
- [x] `node --test benchmarks/*.test.js pi-extension/test/helpers.test.js tests/hooks.test.js tests/openclaw-skills.test.js`
- [x] `npm test`
- [x] `git diff --check`

Closes #120


AGENTS.md
```diff
@@ -25,7 +25,7 @@ Rules:
 - Shortest working diff wins, but only once you understand the problem. The smallest change in the wrong place isn't lazy, it's a second bug.
 - Question complex requests: "Do you actually need X, or does Y cover it?"
 - Pick the edge-case-correct option when two stdlib approaches are the same size, lazy means less code, not the flimsier algorithm.
-- Mark intentional simplifications with a `ponytail:` comment. If the shortcut has a known ceiling (global lock, O(n²) scan, naive heuristic), the comment names the ceiling and the upgrade path.
+- Mark only deliberate simplifications that cut a real corner with a known ceiling (global lock, O(n²) scan, naive heuristic). Normal or trivial code gets no `ponytail:` comment; when a shortcut does need one, the comment names the ceiling and upgrade path.
```
kydorn: this comment applies everywhere. I think that removing the "ponytail:" comment reference is enough. Telling the models to add "no ponytail: comment" is redundant, as there is no longer a reference to adding that ponytail mark

kydorn: in fact, this may be even detrimental. You are basically saying "don't look at the pink elephant" without ever mentioning an elephant before (as you removed the previous elephant reference).

sanmaxdev:
Good point about the confusing reference. Removed the ponytail: comment mention completely to avoid the 'pink elephant' problem.

- git diff --check: passed
- Updated wording is cleaner without the distracting reference

kydorn (approved):

AGENTS.md
```diff
@@ -25,7 +25,7 @@ Rules:
 - Shortest working diff wins, but only once you understand the problem. The smallest change in the wrong place isn't lazy, it's a second bug.
 - Question complex requests: "Do you actually need X, or does Y cover it?"
 - Pick the edge-case-correct option when two stdlib approaches are the same size, lazy means less code, not the flimsier algorithm.
-- Mark only deliberate simplifications that cut a real corner with a known ceiling (global lock, O(n²) scan, naive heuristic). Normal or trivial code gets no `ponytail:` comment; when a shortcut does need one, the comment names the ceiling and upgrade path.
+- Mark deliberate simplifications that cut a real corner with a known ceiling (global lock, O(n²) scan, naive heuristic) with a comment naming the ceiling and upgrade path.
```
kydorn, line 28: you need to apply this same change in ALL the other places.

sanmaxdev:
Updated the copied rule text to match the new wording everywhere.

Checks:
- `npm test`
- `git diff --check`

kydorn (approved):
Lgtm

DietrichGebert:
Thanks for tackling #120, and for the overuse-narrowing, that part was exactly right and I kept it. The one thing I changed my mind on is the `ponytail:` prefix. It's the marker /ponytail-debt greps for, so dropping it would quietly break that skill. So I opened #577, which keeps your narrowing but leaves the prefix in, and also covers the .qoder and .opencode copies that were not around when you opened this. Closing this in favor of that one. Really appreciate the work. Thanks!

PR #494, opened 2026-07-03
docs: add Autohand Code install notes

igorcosta:
## Why

Ponytail already supports a wide set of skill-capable and instruction-only agents. Autohand Code can consume the bundled `SKILL.md` files directly, but the README and portability matrix did not mention the Autohand global or workspace skill paths yet.

## What changed

- Added an Autohand Code install section with direct copy commands for `~/.autohand/skills/` and `.autohand/skills/`.
- Noted that `autohand --skill-install` is for cataloged skills, so users should use the direct copy path until Ponytail is listed there.
- Added Autohand Code to the agent portability table as a skill-capable host.
- Included Autohand Code in the command-capable host sentence.

## Validation

- `git diff --check`
- `npm test` initially showed the documented local `pandas` requirement for CSV checks.
- `PATH="../../.venvs/ponytail-test/bin:$PATH" npm test` passes after installing `pandas` in a workspace venv.


DietrichGebert:
Thanks for the PR, and nice work on Autohand. I am going to pass on this one though. You are the founder and CEO of Autohand, so this is really adding your own product to the README, which is closer to promotion than a neutral contribution, and I keep the host list to tools with broader adoption so it does not turn into an ad board plus translation upkeep across the language READMEs. The genuinely good news: Autohand loads SKILL.md files natively, so ponytail already works on it right now, your users just copy the skills/ folder into ~/.autohand/skills/, no README row needed. If Autohand grows into a widely used host, I am happy to revisit and add it then. Good luck with it.

PR #499, opened 2026-07-03
fix: hermes test hardcoding python3 (fails on Windows)

### LINKED ISSUE #498: tests/hermes-plugin.test.js hardcodes python3 with no python fallback - whole suite fails on Windows
`tests/hermes-plugin.test.js` invokes Python with a hardcoded `python3` and no fallback:

```js
// tests/hermes-plugin.test.js:19
function python(script, env = {}) {
  const result = spawnSync('python3', ['-c', script], { ... });
  if (result.status !== 0) {
    throw new Error(`python failed
STDOUT:...`);
  }
}
```

On a standard Windows machine `python3` resolves to the Microsoft Store app-execution-alias stub, which prints "Python was not found..." and exits non-zero even when Python is installed. Result: every test in the hermes suite that calls `python(...)` throws `Error: python failed`, so the suite fails on Windows despite a working Python.

Reproduced locally: `python` -> `C:\Python314\python`, but `python3` -> the WindowsApps stub, and `node --test tests/hermes-plugin.test.js` errors on every python-backed case.

The repo already solved this exact problem in the benchmark - `benchmarks/correctness.js:49` probes `['python3', 'python']` and falls back:

```js
// ponytail: probe once at load; macOS and many Linux images ship python3 only.
for (const cmd of ['python3', 'python']) { ... }
```

Fix: reuse the same probe/fallback in `hermes-plugin.test.js` instead of hardcoding `python3`.

This is Windows-only and currently invisible in CI because `.github/workflows/test.yml` runs `ubuntu-latest` only (filing that CI gap separately).

trimoyee-g:
## Problem

`tests/hermes-plugin.test.js` spawned `python3` with no fallback. On Windows, `python3` resolves to the Microsoft Store app-execution-alias stub, which exits non-zero even when Python is installed. Every python-backed case in the hermes suite therefore failed on Windows despite a working interpreter.

Reproduced locally: `python` -> `C:\Python314\python`, but `python3` -> the WindowsApps stub, and `node --test tests/hermes-plugin.test.js` errored on every python-backed case.

## Fix

Reuse the probe/fallback pattern already used in `benchmarks/correctness.js`: try `python3`, fall back to `python`, cache the result once.

## Verification

`node --test tests/hermes-plugin.test.js` -> 8 pass, 0 fail (was: every python-backed case erroring on Windows).

Fixes #498

### DIFF
--- a/tests/hermes-plugin.test.js
+++ b/tests/hermes-plugin.test.js
@@ -15,8 +15,21 @@ const skillCommands = commands.filter((name) => name !== 'ponytail');
 
 const root = path.join(__dirname, '..');
 
+// ponytail: probe once; on Windows `python3` is the Store-alias stub that fails
+// even when Python is installed, so fall back to `python` (mirrors benchmarks/correctness.js).
+let pythonCmd;
+function pythonExe() {
+  if (pythonCmd) return pythonCmd;
+  for (const cmd of ['python3', 'python']) {
+    if (spawnSync(cmd, ['-c', 'import sys'], { encoding: 'utf8' }).status === 0) {
+      return (pythonCmd = cmd);
+    }
+  }
+  return (pythonCmd = 'python3');
+}
+
 function python(script, env = {}) {
-  const result = spawnSync('python3', ['-c', script], {
+  const result = spawnSync(pythonExe(), ['-c', script], {
     cwd: root,
     env: { ...process.env, ...env },
     encoding: 'utf8',


DietrichGebert:
Closing: this is byte-identical to #513, which was merged for #498. Both add the python3/python probe mirroring benchmarks/correctness.js. Thanks for the fix.

PR #500, opened 2026-07-03
 fix(pi): recognize and display review mode

capfininv:
## Summary

`review` is already a valid, persisted mode across the project (`VALID_MODES` in `hooks/ponytail-config.js`, referenced by the MCP instructions, the mode-tracker hook, and the Hermes plugin), but the Pi extension hadn't caught up on two fronts:

- The command parser used `normalizeMode` instead of `normalizePersistedMode`, so typing `review` as a command wasn't accepted (it silently fell through as invalid).
- The status bar's `levelIcons` map had no entry for `review`, so a persisted review-mode session rendered with no icon.

## Changes

- `pi-extension/index.js`: accept `review` as a settable mode via the command parser (`normalizePersistedMode`); add a `review: "🔎"` icon so the status bar renders it distinctly from `lite`/`full`/`ultra`.
- Tests covering both: command parsing accepts `review`, and the status bar renders `🔎 REVIEW` for a persisted review-mode session.

## Test plan

- [x] `npm test` — all 77 tests pass (16 root + 61 pi-extension, incl. the 2 new cases)

### DIFF
--- a/pi-extension/index.js
+++ b/pi-extension/index.js
@@ -47,7 +47,7 @@ export function parsePonytailCommand(text, defaultMode = DEFAULT_MODE) {
     return mode ? { type: "set-default", mode } : { type: "invalid", reason: "invalid-default-mode" };
   }
 
-  const mode = normalizeMode(primary);
+  const mode = normalizePersistedMode(primary);
   return mode ? { type: "set-mode", mode } : { type: "invalid", reason: "invalid-mode", mode: primary };
 }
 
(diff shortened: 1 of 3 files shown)


DietrichGebert:
Thanks for this, and the tests are nice. The catch is it makes review a settable, persisted mode, which goes against the current design: review is a one-shot skill (/ponytail-review) and is deliberately kept out of RUNTIME_MODES so it never becomes a session mode or default (#377). On current main normalizePersistedMode rejects review, so the persisted-review path and its icon do not apply, and the branch conflicts now too. Going to close this one, appreciate the effort.

Issue #252, opened 2026-06-22
Ponytail rules don't reach subagents — add SubagentStart hook


halindrome:
## Problem

Ponytail's ruleset is injected by the `SessionStart` hook (`ponytail-activate.js`), which emits it as parent-thread `additionalContext`. Per Claude Code's subagent model, that context does **not** propagate into subagents spawned via the Agent tool — a subagent starts with only its system prompt, CLAUDE.md, memory, and any preloaded skills.

Net effect: ponytail mode governs the main loop but is invisible to every Task-spawned agent. Claude is agentic by design, so the agents doing most of the actual building run ponytail-unaware — which undercuts the point of an always-on mode.

## Fix

Add a `SubagentStart` hook. When ponytail mode is active (read the `.ponytail-active` flag), inject the existing `getPonytailInstructions(mode)` text as `hookSpecificOutput.additionalContext`. No matcher → all agents; ponytail self-limits on non-code agents ("governs what you build, not how you talk"), so inject-into-all is the lazy-correct choice rather than maintaining an agent allowlist.

Note: `writeHookOutput` writes plain stdout on the Claude path (fine for `SessionStart`, not for `SubagentStart`, which needs the `{ hookSpecificOutput: { hookEventName, additionalContext } }` JSON form), so the new hook emits that JSON directly.

## Open question — verify empirically after landing

`SubagentStart` is documented as firing for **Agent-tool** subagents. It is **undocumented** whether:

- Workflow-spawned agents (`agent()` inside Workflow scripts, incl. `parallel()`/`pipeline()` fan-outs), and
- Agent-team members

also fire `SubagentStart`. After the hook lands, verify by spawning an agent via a one-line workflow and a team member, and checking each actually receives the injected ruleset. `TeammateIdle` is an *idle* event, not a start event, so it is not a substitute if teams turn out to be uncovered — that would be a documented limitation, not something to hack around.

## Scope

- New `hooks/ponytail-subagent.js`
- `SubagentStart` entry in `hooks/claude-codex-hooks.json`
- Test mirroring `tests/hooks.test.js`

DietrichGebert:
Merged. Verified the hook live first: spawned an Agent-tool subagent with the SubagentStart hook wired in and it received the full ruleset verbatim, a control run without the hook got nothing. So the injection is real and the hook is what delivers it, not something leaking in from CLAUDE.md or memory.

[...]

PR #507, opened 2026-07-04
fix: add agent_type matcher to SubagentStart hook (#506)

### LINKED ISSUE #506: SubagentStart hook has no matcher — allow scoping the persona injection to specific agent types
## Request

Add an **agent-type matcher (or opt-in scoping config)** to the plugin's `SubagentStart` hook so consumers can limit the persona injection to specific subagent types.

## Why

The `SubagentStart` hook currently ships with no matcher, so its ~1,300-token persona is injected into **every** subagent spawned via the Agent tool — including cheap read-only agents (search/verify) and low-tier agents whose entire base prompt is ~150–200 words. On those it's a 5–6× prompt tax and, for agents that never write code, dead weight.

Claude Code `SubagentStart` hooks support a `matcher` that filters by agent type name, and the hook receives `agent_type` on stdin — so scoping is technically possible. The problem is it can only be done **on the hook's own definition**: a consumer cannot suppress or narrow a plugin-provided hook from their own `settings.json` (plugin hooks merge with user/project hooks; both fire). So the only place this can be fixed is upstream, here.

1. Add a default `matcher` on the SubagentStart hook that targets code-writing agents, or
2. Make the matcher configurable via plugin settings (e.g. an `agents` allowlist), or
3. Have the hook self-filter on the `agent_type` it receives on stdin against a documented default/env-overridable list.

Opt-in scoping (off by default = current behavior) would be the least disruptive.

## Context

Filed from a downstream config repo where this surfaced in a per-call overhead audit. Happy to send a PR if you point me at the preferred approach.

dhedhialy:
Adds `PONYTAIL_SUBAGENT_MATCHER` env var support to the SubagentStart hook. When set, the hook reads `agent_type` from stdin and only injects ponytail instructions when the agent type matches the regex. Opt-in by default (no env var = inject all subagents, preserving backward compatibility).

Fixes #506

### DIFF
--- a/hooks/ponytail-subagent.js
+++ b/hooks/ponytail-subagent.js
@@ -4,6 +4,11 @@
 // SessionStart context is parent-thread only and never reaches subagents, so
 // without this every Task-spawned agent runs ponytail-unaware (issue #252).
 // When ponytail mode is active, inject the same ruleset into each subagent.
+//
+// Scoping: by default injects into all subagents. To limit to specific agent
+// types, set PONYTAIL_SUBAGENT_MATCHER to a regex (e.g. "explore|general").
+// The hook reads agent_type from stdin and skips injection when it doesn't
+// match. This is opt-in to preserve backward compatibility (#506).
 
 const { getPonytailInstructions } = require('./ponytail-instructions');
 const { readMode, writeHookOutput } = require('./ponytail-runtime');
@@ -15,8 +20,32 @@ if (!mode || mode === 'off') {
   process.exit(0);
 }
 
-try {
-  writeHookOutput('SubagentStart', mode, getPonytailInstructions(mode));
-} catch (e) {
-  // Silent fail — a stdout error at hook exit must not surface as a hook failure.
-}
+// ponytail: opt-in scoping via env var. Off by default = all subagents.
+// Set PONYTAIL_SUBAGENT_MATCHER to a regex matching the agent types to inject.
+// Upgrade path: if many users set this, promote to config.json.
+const matcher = process.env.PONYTAIL_SUBAGENT_MATCHER;
+const matcherRe = matcher ? new RegExp(matcher, 'i') : null;
+
+let input = '';
+
+process.stdin.on('data', chunk => { input += chunk; });
+process.stdin.on('end', () => {
+  if (matcherRe) {
+    try {
+      const data = JSON.parse(input.replace(/^\uFEFF/, ''));
+      const agentType = (data.agent_type || data.agentType || '').trim();
+      if (!matcherRe.test(agentType)) process.exit(0);
+    } catch (_) {
+      // Can't parse input or missing agent_type — inject to be safe.
+    }
+  }
+  try {
+    writeHookOutput('SubagentStart', mode, getPonytailInstructions(mode));
+  } catch (e) {
+    // Silent fail — a stdout error at hook exit must not surface as a hook failure.
+  }
+});
+
+// Mirror the safety timeout from mode-tracker so the hook never blocks.
+process.stdin.on('error', () => { process.exit(0); });
+setTimeout(() => { process.exit(0); }, 1000).unref();
--- a/hooks/ponytail-subagent.js
+++ b/hooks/ponytail-subagent.js
@@ -24,28 +24,49 @@ if (!mode || mode === 'off') {
 // Set PONYTAIL_SUBAGENT_MATCHER to a regex matching the agent types to inject.
 // Upgrade path: if many users set this, promote to config.json.
 const matcher = process.env.PONYTAIL_SUBAGENT_MATCHER;
-const matcherRe = matcher ? new RegExp(matcher, 'i') : null;
+let matcherRe = null;
+try {
+  if (matcher) matcherRe = new RegExp(matcher, 'i');
+} catch (_) {
+  // Invalid regex must not crash the hook — fall back to injecting everywhere.
+}
 
-let input = '';
-
-process.stdin.on('data', chunk => { input += chunk; });
-process.stdin.on('end', () => {
-  if (matcherRe) {
-    try {
-      const data = JSON.parse(input.replace(/^\uFEFF/, ''));
-      const agentType = (data.agent_type || data.agentType || '').trim();
-      if (!matcherRe.test(agentType)) process.exit(0);
-    } catch (_) {
-      // Can't parse input or missing agent_type — inject to be safe.
-    }
-  }
+function inject() {
   try {
     writeHookOutput('SubagentStart', mode, getPonytailInstructions(mode));
   } catch (e) {
     // Silent fail — a stdout error at hook exit must not surface as a hook failure.
   }
-});
+}
 
-// Mirror the safety timeout from mode-tracker so the hook never blocks.
-process.stdin.on('error', () => { process.exit(0); });
-setTimeout(() => { process.exit(0); }, 1000).unref();
+// No matcher: keep the original synchronous, stdin-independent behavior. On
+// Windows the PowerShell `if {}` wrapper can swallow stdin so 'end' never fires
+// (#443), so the default path must not depend on reading stdin.
+if (!matcherRe) {
+  inject();
+  process.exit(0);
+}
+
+// Matcher set: read agent_type from stdin, skip injection only on a definite
+// non-match. Fail OPEN (inject) on unparseable input, stdin error, or timeout.
+let input = '';
+let done = false;
+
+function finish() {
+  if (done) return;
+  done = true;
+  try {
+    const data = JSON.parse(input.replace(/^\uFEFF/, ''));
+    const agentType = (data.agent_type || '').trim();
+    if (agentType && !matcherRe.test(agentType)) process.exit(0);
+  } catch (_) {
+    // Can't parse input or missing agent_type — inject to be safe.
+  }
+  inject();
+}
+
+process.stdin.on('data', chunk => { input += chunk; });
+process.stdin.on('end', finish);
+// Mirror mode-tracker's never-block contract: recover on error/timeout too.
+process.stdin.on('error', () => { finish(); process.exit(0); });
+setTimeout(() => { finish(); process.exit(0); }, 1000).unref();
--- a/tests/hooks.test.js
+++ b/tests/hooks.test.js
@@ -215,46 +215,31 @@ assert.equal(output.systemMessage, 'PONYTAIL:FULL');
 assert.equal(output.hookSpecificOutput.hookEventName, 'SubagentStart');
 assert.match(output.hookSpecificOutput.additionalContext, /PONYTAIL MODE ACTIVE — level: full/);
 
-// writeDefaultMode must merge into existing config, not overwrite it (#490).
-const mergeHome = path.join(temp, 'merge-home');
-const mergeConfigDir = path.join(mergeHome, '.config', 'ponytail');
-fs.mkdirSync(mergeConfigDir, { recursive: true });
-const mergeConfigPath = path.join(mergeConfigDir, 'config.json');
-fs.writeFileSync(mergeConfigPath, JSON.stringify({ defaultMode: 'full', customSetting: 42 }, null, 2));
+// SubagentStart matcher (issue #506): PONYTAIL_SUBAGENT_MATCHER filters by agent_type.
+const matcherHome = path.join(temp, 'matcher-home');
+const matcherFlag = path.join(matcherHome, '.claude', '.ponytail-active');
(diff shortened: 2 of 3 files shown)


dhedhialy:

hooks/ponytail-subagent.js
```diff
@@ -15,8 +20,32 @@ if (!mode || mode === 'off') {
...
+  try {
+    writeHookOutput('SubagentStart', mode, getPonytailInstructions(mode));
+  } catch (e) {
+    // Silent fail — a stdout error at hook exit must not surface as a hook failure.
+  }
+});
+
+// Mirror the safety timeout from mode-tracker so the hook never blocks.
+process.stdin.on('error', () => { process.exit(0); });
+setTimeout(() => { process.exit(0); }, 1000).unref();
```
dividedby: **Two blocking issues, fixable together in this block.**

**1. Windows regression (#443 class).** The base hook wrote output synchronously and never read stdin, so it always injected. Moving `writeHookOutput` inside `stdin.on('end', …)` makes the *default* (no-matcher) path depend on `end` firing. `ponytail-mode-tracker.js` documents that on Windows the PowerShell `if {}` wrapper swallows the piped JSON so `end` never fires — and it recovers by calling `finish()` from its timeout/error handlers. This hook copied the timeout's *shape* but its `error`/timeout handlers `process.exit(0)` **without writing**, so on Windows the default path silently stops injecting — regressing the very thing the hook exists for (#252).

**2. Invalid regex crashes the hook.** `new RegExp(matcher, 'i')` (L27) runs at module top level with no guard. `PONYTAIL_SUBAGENT_MATCHER='['` throws → non-zero exit → hook failure on *every* subagent spawn, contradicting the "must not surface as a hook failure" contract on L45.

Fix: keep the no-matcher path synchronous and stdin-independent (byte-for-byte today's behavior — no Windows regression, no added latency), guard the regex, and read stdin only when a matcher is set — failing **open** (inject) on parse error / stdin error / timeout, matching the stated intent.

```suggestion
// ponytail: opt-in scoping via env var. Off by default = all subagents.
// Set PONYTAIL_SUBAGENT_MATCHER to a regex matching the agent types to inject.
// Upgrade path: if many users set this, promote to config.json.
const matcher = process.env.PONYTAIL_SUBAGENT_MATCHER;
let matcherRe = null;
try {
  if (matcher) matcherRe = new RegExp(matcher, 'i');
} catch (_) {
  // Invalid regex must not crash the hook — fall back to injecting everywhere.
}

function inject() {
  try {
    writeHookOutput('SubagentStart', mode, getPonytailInstructions(mode));
  } catch (e) {
    // Silent fail — a stdout error at hook exit must not surface as a hook failure.
  }
}

// No matcher: keep the original synchronous, stdin-independent behavior. On
// Windows the PowerShell `if {}` wrapper can swallow stdin so 'end' never fires
// (#443), so the default path must not depend on reading stdin.
if (!matcherRe) {
  inject();
  process.exit(0);
}

// Matcher set: read agent_type from stdin, skip injection only on a definite
// non-match. Fail OPEN (inject) on unparseable input, stdin error, or timeout.
let input = '';
let done = false;

function finish() {
  if (done) return;
  done = true;
  try {
    const data = JSON.parse(input.replace(/^\uFEFF/, ''));
    const agentType = (data.agent_type || '').trim();
    if (agentType && !matcherRe.test(agentType)) process.exit(0);
  } catch (_) {
    // Can't parse input or missing agent_type — inject to be safe.
  }
  inject();
}

process.stdin.on('data', chunk => { input += chunk; });
process.stdin.on('end', finish);
// Mirror mode-tracker's never-block contract: recover on error/timeout too.
process.stdin.on('error', () => { finish(); process.exit(0); });
setTimeout(() => { finish(); process.exit(0); }, 1000).unref();
```

tests/hooks.test.js
```diff
@@ -209,4 +209,31 @@ assert.equal(output.systemMessage, 'PONYTAIL:FULL');
...
+  JSON.stringify({ agent_type: 'explore' }));
+assert.equal(result.status, 0, result.stderr);
+assert.equal(result.stdout, '', 'should skip injection when agent_type does not match');
+
+// No matcher env → inject regardless of agent_type.
+result = run('ponytail-subagent.js', matcherEnv,
+  JSON.stringify({ agent_type: 'explore' }));
+assert.equal(result.status, 0, result.stderr);
+output = JSON.parse(result.stdout);
+assert.ok(output.hookSpecificOutput, 'should inject when no matcher is set');
```
dividedby, line 245: Add coverage for the two fail-open paths — the first case would have caught the top-level `new RegExp` crash:

```suggestion
// No matcher env → inject regardless of agent_type.
result = run('ponytail-subagent.js', matcherEnv,
  JSON.stringify({ agent_type: 'explore' }));
assert.equal(result.status, 0, result.stderr);
output = JSON.parse(result.stdout);
assert.ok(output.hookSpecificOutput, 'should inject when no matcher is set');

// Invalid regex → must not crash the hook; fail open (inject).
result = run('ponytail-subagent.js', { ...matcherEnv, PONYTAIL_SUBAGENT_MATCHER: '[' },
  JSON.stringify({ agent_type: 'explore' }));
assert.equal(result.status, 0, result.stderr);
output = JSON.parse(result.stdout);
assert.ok(output.hookSpecificOutput, 'invalid matcher regex should inject, not crash');

// Matcher set but agent_type absent → inject to be safe.
result = run('ponytail-subagent.js', { ...matcherEnv, PONYTAIL_SUBAGENT_MATCHER: 'general' },
  JSON.stringify({}));
assert.equal(result.status, 0, result.stderr);
output = JSON.parse(result.stdout);
assert.ok(output.hookSpecificOutput, 'absent agent_type should inject to be safe');
```

dividedby (changes requested):
Reviewed the mechanism end to end. `agent_type` is the correct SubagentStart stdin field (confirmed against the Claude Code hooks docs), so the matching approach is sound. Requesting changes on two correctness issues introduced by the sync→async rewrite (inline), plus one docs gap.

**Docs — definition of done.** `PONYTAIL_SUBAGENT_MATCHER` is user-facing but undocumented. Its sibling `PONYTAIL_DEFAULT_MODE` is documented in `README.md` (~L248 and the FAQ ~L301) — add `PONYTAIL_SUBAGENT_MATCHER` alongside it, noting it's an **unanchored, case-insensitive** regex tested against `agent_type` (use `^…$` for an exact match; plugin agent types look like `plugin:name`). No CHANGELOG file exists in the repo, so nothing to update there.

The inline comments carry directly-applyable suggestions for the two code fixes.

dhedhialy:
Addressed all review feedback:\n\n1. **Windows regression**: no-matcher path is now synchronous and stdin-independent (identical to pre-PR behavior)\n2. **Invalid regex**:  is guarded by try/catch — falls back to injecting everywhere\n3. **Documentation**:  documented in README.md alongside \n4. **Test coverage**: added invalid-regex and absent-agent_type cases (fail-open)

dhedhialy:
@dividedby all feedback addressed — the three changes are pushed to the same branch. Ready for re-review.

dividedby (approved):
Re-reviewed the pushed fixup (`c167f04`) against all four items from the prior `CHANGES_REQUESTED`. All addressed correctly — approving.

- **Windows regression (#443 class) — fixed.** The no-matcher path is now `if (!matcherRe) { inject(); process.exit(0); }` — synchronous and stdin-independent, so the default path no longer depends on `end` firing. The `process.exit(0)` there is load-bearing (prevents fall-through into the stdin block). Verified no truncation: the no-matcher test JSON-parses the full ~5.3 KB payload through a pipe and passes.
- **Invalid-regex crash — fixed.** `new RegExp(matcher, 'i')` is now guarded by try/catch; an invalid pattern leaves `matcherRe` null and falls into the inject-everywhere path. Covered by the `'['` test case.
- **Docs — fixed.** `PONYTAIL_SUBAGENT_MATCHER` is documented in `README.md` alongside `PONYTAIL_DEFAULT_MODE`, noting it's an unanchored, case-insensitive regex against `agent_type`, with `^…$` for exact match and the `plugin:name` caveat.
- **Tests — fixed.** Fail-open coverage added (invalid regex, absent `agent_type`) on top of the existing positive/negative-match cases. Full suite passes locally (`node tests/hooks.test.js`).

Matcher-path fail-open logic is correct: injects on unparseable input, empty/absent `agent_type`, stdin error, and timeout — skips only on a definite non-match, mirroring `ponytail-mode-tracker.js`'s never-block contract. Dropping the base commit's `data.agentType` camelCase fallback is also correct — `agent_type` is the real SubagentStart stdin field.

Nice work.

dhedhialy:
This PR has been approved by @dividedby. Could someone with write access merge it?

tests/hooks.test.js
```diff
@@ -209,4 +209,45 @@ assert.equal(output.systemMessage, 'PONYTAIL:FULL');
...
+assert.equal(result.status, 0, result.stderr);
+output = JSON.parse(result.stdout);
+assert.ok(output.hookSpecificOutput, 'invalid matcher regex should inject, not crash');
+
+// Matcher set but agent_type absent → inject to be safe.
+result = run('ponytail-subagent.js', { ...matcherEnv, PONYTAIL_SUBAGENT_MATCHER: 'general' },
+  JSON.stringify({}));
+assert.equal(result.status, 0, result.stderr);
+output = JSON.parse(result.stdout);
+assert.ok(output.hookSpecificOutput, 'absent agent_type should inject to be safe');
```
dividedby, line 259: Ported from #522 (verified locally, suite green). Adds case-insensitive/unanchored match, anchored-regex rejection, and the empty-stdin default-path check.

```suggestion
assert.ok(output.hookSpecificOutput, 'absent agent_type should inject to be safe');

// Case-insensitive, unanchored match → inject ('General-purpose' matches 'general').
result = run('ponytail-subagent.js', { ...matcherEnv, PONYTAIL_SUBAGENT_MATCHER: 'general|plan' },
  JSON.stringify({ agent_type: 'General-purpose' }));
assert.equal(result.status, 0, result.stderr);
output = JSON.parse(result.stdout);
assert.ok(output.hookSpecificOutput, 'matcher should be case-insensitive and unanchored');

// Anchored regex → exact match only; a superset agent_type is rejected.
result = run('ponytail-subagent.js', { ...matcherEnv, PONYTAIL_SUBAGENT_MATCHER: '^general$' },
  JSON.stringify({ agent_type: 'general-purpose' }));
assert.equal(result.status, 0, result.stderr);
assert.equal(result.stdout, '', 'anchored matcher must not match a superset agent_type');

// No matcher + empty stdin → default path injects without waiting on stdin (#443).
result = run('ponytail-subagent.js', matcherEnv, '');
assert.equal(result.status, 0, result.stderr);
output = JSON.parse(result.stdout);
assert.ok(output.hookSpecificOutput, 'no-matcher path must not depend on stdin');
```

dividedby (commented):
Porting the extra test coverage from #522, as discussed. I applied both additions to this branch locally and ran the full suite — green on Node 26.

**1. Three matcher cases** — inline suggestion below (one-click apply). They cover gaps the current tests don't exercise:
- case-insensitive + unanchored match (`General-purpose` against `general|plan`)
- anchored `^general$` rejecting the superset `general-purpose` (proves the "use `^…$` for exact" README note actually holds)
- no-matcher path with **empty stdin** — the most direct assertion that the default path never waits on stdin (the #443 Windows fix)

**2. One harness-hygiene line** — this one doesn't sit next to a diff hunk, so it can't be a suggestion; add it by hand in the `delete process.env.*` block near the top of the file (right after `delete process.env.COPILOT_PLUGIN_DATA;`):

```js
// A leaked matcher would scope the inject-into-every-subagent assertions.
delete process.env.PONYTAIL_SUBAGENT_MATCHER;
```

Without it, a `PONYTAIL_SUBAGENT_MATCHER` set in the CI/runner environment would leak through `run()` (which spreads `process.env`) and silently scope the earlier inject-into-every-subagent assertions.

Approval stands regardless — this is additive coverage, not a blocker.

dhedhialy:
Ported the additional coverage from #522 as suggested: case-insensitive match, anchored rejection, empty-stdin default path test, and the harness-hygiene delete. All 85 tests green.

dhedhialy:

DietrichGebert:
Nice work on the subagent scoping. We have #522 doing the same agent_type matcher, and it stacks on current main without dropping the #490 and #329 test coverage that this branch removes, so I am going with that one. Thanks a lot for jumping on #506.

dhedhialy:
Rebased and all conflicts resolved — now MERGEABLE. All tests pass (90/90 including pi-extension). Ready for review when you have time.

dhedhialy:
Understood — #522 is the cleaner approach. Closing this one. Thanks for the explanation.

PR #509, opened 2026-07-04
fix: remove .openclaw/skills/ to fix Copilot CLI duplicate skill collisions (#501)

### LINKED ISSUE #501: bug: Copilot CLI reports 6 duplicate skill failures from .openclaw/skills/
## Title
bug: Copilot CLI reports 6 duplicate skill failures from .openclaw/skills/
### Problem
When ponytail is installed as a **GitHub Copilot CLI** plugin, the CLI reports 6 skill loading failures at every startup:
```
Failed to load 6 skills. Run /skills for more details
```
**Root cause:** Copilot CLI recursively discovers every `SKILL.md` file inside the installed plugin directory. It finds skills in both `skills/` and `.openclaw/skills/`, which have identical `name:` values — 6 collisions.
```
FAIL: ponytail         (.openclaw/skills/ponytail        collides with skills/ponytail)
FAIL: ponytail-audit    (.openclaw/skills/ponytail-audit   collides with skills/ponytail-audit)
FAIL: ponytail-debt     (.openclaw/skills/ponytail-debt    collides with skills/ponytail-debt)
FAIL: ponytail-gain     (.openclaw/skills/ponytail-gain    collides with skills/ponytail-gain)
FAIL: ponytail-help     (.openclaw/skills/ponytail-help    collides with skills/ponytail-help)
FAIL: ponytail-review   (.openclaw/skills/ponytail-review  collides with skills/ponytail-review)
```
The plugin correctly declares `"skills": "skills/"` in `.github/plugin/plugin.json`, but the CLI ignores that manifest and scans all directories.
### Analysis
`.openclaw/skills/` has shorter, OpenClaw-format descriptions and extra frontmatter fields (`homepage`, `license`). It is clearly intended for a different tool and should not be loaded by Copilot CLI.
The `package.json` `"files"` field does **not** include `.openclaw/`, so npm packages exclude it correctly. However, Copilot CLI installs via `copilot plugin marketplace add` + `copilot plugin install`, which clones the full repo — `.openclaw/` ends up on disk even though it is not declared as a Copilot CLI asset.
### Why this matters
Every Copilot CLI startup shows a noisy failure banner. Users who install ponytail alongside other multi-tool plugins can hit 20+ phantom failures, making it impossible to tell whether a real loading error exists.
### Proposed solutions
Any of these would fix it from the plugin side:
| # | Approach | Effort | Side-effects |
|---|---|---|---|
| 1 | **Deduplicate**: drop `.openclaw/skills/` entirely and point OpenClaw at the existing `skills/` directory. Both use SKILL.md format; differences are just description length + 2 extra frontmatter keys (`homepage`, `license`). | Low | OpenClaw gets the longer descriptions — arguably better |
| 2 | **Rename `.openclaw/skills/*/SKILL.md`** to a name OpenClaw supports but Copilot CLI does not scan (e.g. `skill.md` lowercase). | Low | Only works if OpenClaw accepts alternative filenames |
| 3 | **Exclude `.openclaw/` from the git-cloned copy** that Copilot CLI installs. Use `.gitattributes export-ignore` so `git archive` (and tools that use it) skip `.openclaw/`. | Medium | Depends on how Copilot CLI clones the repo |

Option 1 is simplest and removes the maintenance burden of keeping two directories in sync.
From the **Copilot CLI side**, the CLI should respect `.github/plugin/plugin.json` manifests and only scan declared directories. This is arguably a CLI bug, but the workaround above is fully in the plugin author's control.
### Repro
```bash
copilot plugin marketplace add DietrichGebert/ponytail
copilot plugin install ponytail@ponytail
copilot   # → "Failed to load 6 skills"
```
### Environment
- **Copilot CLI** latest
- **ponytail** v4.8.4
- macOS (also likely affects Linux)
- `.github/plugin/plugin.json` present with `"skills": "skills/"`

dhedhialy:
Copilot CLI recursively discovers every SKILL.md inside the repo, finding duplicates in both skills/ and .openclaw/skills/. Removed the duplicate .openclaw/skills/ directory, deleted the now-unnecessary generator/publish scripts and stale test. OpenClaw users can use skills/ directly.

Fixes #501


dhedhialy:

dhedhialy:

DietrichGebert:
Thanks for digging into the Copilot CLI collision. The thing is, .openclaw/skills/ is a channel we publish to ClawHub on purpose, so removing the whole pipeline takes out more than we want and leaves the Spanish and Korean READMEs pointing at files that are gone. Better to fix it on the Copilot side so it stops reading that folder. Closing this one, but the bug report really helps.

dhedhialy:
Rebased and all conflicts resolved — now MERGEABLE. All tests pass (90/90 including pi-extension). Ready for review when you have time.

dhedhialy:
Makes sense — the ClawHub pipeline is the intended user of .openclaw/skills/. Closing this. Thanks for the context.

PR #510, opened 2026-07-04
fix: use valid model gpt-4o-mini as default in robustness-audit.js (#399)

### LINKED ISSUE #399: benchmarks/robustness-audit.js: hardcodes invalid model gpt-5.4-mini as default

dhedhialy:
The default model `gpt-5.4-mini` is not a valid OpenAI model ID. Changed to `gpt-4o-mini`. Can be overridden with `AUDIT_MODEL` env var.

Fixes #399

### DIFF
--- a/benchmarks/robustness-audit.js
+++ b/benchmarks/robustness-audit.js
@@ -2,7 +2,7 @@
 // weak model. 12 tasks with classic edge-case traps. Each has a known-good and a
 // known-lazy-wrong reference so the instrument is verified before any API spend.
 //   node robustness-audit.js --selftest   # no API: prove every check is correct
-//   node robustness-audit.js              # baseline vs ponytail, gpt-5.4-mini, n=20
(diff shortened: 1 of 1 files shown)


dhedhialy:

DietrichGebert:
Thanks @dhedhialy, but closing without merging. gpt-5.4-mini is a valid, current OpenAI model (callable alias; snapshot gpt-5.4-mini-2026-03-17), described in OpenAI docs as their strongest mini model: https://developers.openai.com/api/docs/models/gpt-5.4-mini

So #399 is invalid, and this change would swap a working, current default for an older model (gpt-4o-mini) without fixing anything. The GPT-5.4 series shipped March 2026, so gpt-5.4-mini likely just looked unfamiliar. The benchmark also already logs real results on gpt-5.4-mini, so it is the intended default. Appreciate the PR regardless.

dhedhialy:

PR #515, opened 2026-07-04

### LINKED ISSUE #329: pi ponytail on and off issue

dhedhialy:
The mode-tracker now calls writeDefaultMode() on every mode change, so /ponytail off is persisted to config.json and survives session restart.

Fixes #329

### DIFF
--- a/hooks/ponytail-mode-tracker.js
+++ b/hooks/ponytail-mode-tracker.js
@@ -2,7 +2,7 @@
 // ponytail — UserPromptSubmit hook to track which ponytail mode is active
 // Inspects user input for /ponytail commands and writes mode to flag file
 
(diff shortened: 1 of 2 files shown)


dhedhialy:

dhedhialy:

DietrichGebert:
Closing in favor of #541 (merged), which took a different approach. Persisting every /ponytail switch (as this PR does) contradicts the documented session-scoped behavior ("Level sticks until changed or session end") and would turn a one-off /ponytail ultra into a permanent default. #541 instead adds an explicit /ponytail default <mode> command (matching pi) that persists the default while keeping plain switches transient. Thanks for the fix.

PR #518, opened 2026-07-04
fix: handle CRLF in loc.js fence regex and remove dead codeOf (#339)

### LINKED ISSUE #339: benchmarks: loc.js misses CRLF code fences and counts the whole response

dhedhialy:
loc.js regex was missing \r? before \n, so CRLF code fences were never matched and the metric fell back to counting the whole response. Also removed dead codeOf() from behavior.js (defined but never called).

Fixes #339

### DIFF
--- a/benchmarks/behavior.js
+++ b/benchmarks/behavior.js
@@ -10,10 +10,6 @@
 //
 // Metric: `behavior` (1 = behavior present, 0 = absent).
 
-function codeOf(text) {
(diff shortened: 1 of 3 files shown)


dhedhialy:

dhedhialy:

PR #519, opened 2026-07-04
fix: use computed settingsPath in statusline nudge (#250)

### LINKED ISSUE #250: Statusline setup nudge hardcodes ~/.claude/settings.json, ignoring CLAUDE_CONFIG_DIR

dhedhialy:
When CLAUDE_CONFIG_DIR is set, the statusline setup nudge was telling users to edit ~/.claude/settings.json — the wrong file. Now interpolates the already-computed settingsPath so the nudge points at the right file.

Fixes #250


dhedhialy:

dhedhialy:

DietrichGebert:
Closing in favor of #338 (merged), which is the same settingsPath interpolation fix for #250. I took #338 because its regression test is unconditional (it can not pass vacuously if the nudge does not fire). Thanks for the fix.

PR #522, opened 2026-07-05
feat(hooks): opt-in agent-type scoping for SubagentStart injection (#…

### LINKED ISSUE #506: SubagentStart hook has no matcher — allow scoping the persona injection to specific agent types

krishrathi1:
…506)

The SubagentStart hook injected the ~1,300-token persona into every subagent, including read-only search agents where it is dead weight. Consumers could not narrow a plugin-provided hook from their own settings, so the scoping has to live here.

Add an opt-in allowlist: PONYTAIL_SUBAGENT_AGENTS (comma-separated env var) or a subagentAgents array in config.json, matching getDefaultMode precedence. The hook now reads agent_type from stdin and skips injection only on a positive mismatch; unconfigured or missing agent_type keeps the inject-everywhere behavior from issue #252. Stdin reading follows the never-hang contract from issue #443.

Closes #506


dhedhialy (changes requested):
## Blocking issues

**1. Windows regression (#443 class).** `writeHookOutput` now lives inside `stdin.on(end, finish)` on every code path — including the default (no allowlist) path. The mode-tracker hook documents that on Windows the PowerShell `if {}` wrapper swallows the piped JSON so `end` never fires. Your `setTimeout` will eventually fire at 1s, but the hook is not writing anything until then. On the default path (no `PONYTAIL_SUBAGENT_AGENTS` set) the hook used to inject synchronously with no stdin dependency. Now it injects only after a 1s timeout or never. This is a regression of #252 and contradicts the "never block" contract.

**2. `finish()` returns without `process.exit(0)` on the skip path.** When the agent type is outside the allowlist, you `return` from `finish()` without calling `process.exit(0)`. The process stays alive until the 1s `setTimeout` fires, adding 1s of latency to every skipped subagent spawn. The error- and timeout-path handlers call `finish(); process.exit(0)` — the `finish();` call on line end would skip (correct), then `process.exit(0)` runs. But `stdin.on(end, finish)` has no `process.exit(0)` after it, so on the normal path (end fires) a skip leaves the process lingering.

**3. `getSubagentAgents()` called per-spawn instead of once at module load.** It reads the config file from disk every time a subagent spawns. `getDefaultMode` is also called per-spawn elsewhere but that is cached by the OS page cache — fair. Adding config-file I/O to the subagent hot path when the result is trivially hoistable is unnecessary.

## Design issues

**4. Comma-separated list over a regex.** `PONYTAIL_SUBAGENT_AGENTS` uses exact string matching (`allowed.includes(...)`). A regex matcher is more flexible (anchored `^general$` for exact, `explore|general` for union, `plugin:` prefix matching) and requires zero pluralization/trim/normalization code. The regex approach was already reviewed and approved on PR #507 — this is a different API surface for the same feature.

**5. Env var naming.** `PONYTAIL_SUBAGENT_AGENTS` is inconsistent with the established `PONYTAIL_DEFAULT_MODE` pattern. Both are env+config precedences; having `_MODE` vs `_AGENTS` makes discovery harder. `PONYTAIL_SUBAGENT_MATCHER` (or `_FILTER`, `_ALLOWLIST`) would match the `_MODE` convention better.

**6. Unrelated test file modification.** `tests/hooks-windows.test.js` added a test for the subagent freeze, but this file is specifically for Windows-specific behavior. The freeze test duplicates what could be tested once. It also imports `os` at module top-level in the test body rather than at the top of the file.

## Nits

- Import `os` at the top of hooks-windows.test.js alongside the other imports, not inside the test body.
- The `done` guard pattern works but without `process.exit(0)` on the skip path it leaks processes.

**tl;dr:** The fundamental issues are: (a) the default path regresses Windows by depending on stdin, (b) the skip path leaks 1s per subagent spawn, and (c) the per-call config read is wasteful. These are the same class of bugs @dividedby caught in PR #507.

krishrathi1 (commented):
i had updated and resolved the issues

dividedby:
Flagging a duplicate: this and #507 both `Closes #506`, and after the "Address review: regex matcher…" commit here, the two have converged on the **same** design — the `PONYTAIL_SUBAGENT_MATCHER` env var, an unanchored/case-insensitive regex against `agent_type`, a synchronous stdin-independent default path (#443), an invalid-regex guard, fail-open on missing/unparseable `agent_type` and timeout, and the same README slot. Same env var name, same semantics, same three files.

#507 was opened earlier and is already reviewed and approved (merge-ready), so it makes sense to consolidate there and close this one to avoid a merge conflict over the identical env var and hook.

That said, the tests here are a bit more thorough than #507's, and are worth porting over before #507 merges:
- case-insensitive match assertion (`General-purpose` against `general|plan`);
- anchored `^general$` rejecting the superset `general-purpose`;
- empty-stdin default path (proves the no-matcher path never waits on stdin);
- `delete process.env.PONYTAIL_SUBAGENT_MATCHER` in the harness setup to stop a leaked matcher from scoping the inject-everywhere assertions.

Nice work — the implementation is solid; it's purely a timing/duplication call. Deferring to @DietrichGebert on which to keep.

dhedhialy (approved):

krishrathi1:
Hi @DietrichGebert, the PR has no merge conflicts but the workflow is awaiting maintainer approval. Could you please approve/run the workflow and merge it if the checks pass?

krishrathi1:
Hi @DietrichGebert, the PR has no merge conflicts but the workflow is awaiting maintainer approval. Could you please approve/run the workflow and merge it if the checks pass?

Issue #461, opened 2026-06-30
Bug: Codex plugin lifecycle hooks (ponytail-mode-tracker.js) trap shell in an infinite background fork loop on low maxuserproc machines


Vaibhav159:
### Title
`bug: Codex plugin lifecycle hooks (ponytail-mode-tracker.js) trap shell in an infinite background fork loop on low maxuserproc machines`

### Description
When installing Ponytail inside the Codex CLI ecosystem, the background execution scripts (`ponytail-mode-tracker.js` and `ponytail-activate.js`) get stuck in an unhandled, infinite loop of spawning sub-processes. 

Every single interactive prompt or Zsh hook execution triggers a wall of zombie sub-processes that never close, immediately hitting the system's `maxuserproc` limit (especially on Apple Silicon Macs where default process limits are set to ~1064). 

Furthermore, `codex plugin remove` fails to unbind the active shell hooks cleanly if a marketplace format error occurs, trapping the terminal in an un-killable loop where killed PIDs are instantly re-spawned by parent background triggers.

### Steps to Reproduce
1. Install Ponytail in Codex via `codex plugin marketplace add` and follow the instructions to trust shell lifecycle hooks.
2. Open a standard `zsh` session or an integrated terminal pane (e.g., inside IntelliJ/WebStorm).
3. Type normal command variations (e.g., `git status` or trigger auto-suggestions).
4. Run `ps aux | grep ponytail`.

### Expected Behavior
The plugin should spawn single-use hooks, or cleanly terminate active tracking processes when the parent prompt hook completes or the plugin is deactivated.

### Actual Behavior
The terminal floods with `fork failed: resource temporarily unavailable` across all packages (Homebrew, P10k, internal shell, etc.), and process tracing shows an endless, compounding list of hanging shell calls:

```text
user    45968  16.9  0.0 435448320   4192   ??  S      7:59PM   0:00.07 /bin/zsh -c node "/Users/user/.codex/plugins/cache/ponytail/ponytail/4.8.3/hooks/ponytail-mode-tracker.js"; exit 0
user   45893  11.1  0.0 435442912   4224   ??  S      7:59PM   0:00.07 /bin/zsh -c node "/Users/user/.codex/plugins/cache/ponytail/ponytail/4.8.3/hooks/ponytail-mode-tracker.js"; exit 0
[...]

PR #529, opened 2026-07-06
feat: add Qoder hooks support (UserPromptSubmit + PreToolUse)

shixiaoding:
- Add isQoder detection via QODER_SESSION_ID in ponytail-runtime.js
- Add UserPromptSubmit double-duty: auto-activate + ruleset injection
- Add PreToolUse hook with task|Task matcher for subagent injection
- Create hooks/qoder-hooks.json reference template
- Update .qoder-plugin/plugin.json with hooks field
- Add .qoder/rules/ponytail.md for per-project rules
- Add comprehensive tests for Qoder runtime behavior
- Update README.md and docs/agent-portability.md

PR #530, opened 2026-07-06
fix: restore VS Code Copilot hook compatibility on Windows

### LINKED ISSUE #527: VS Code Copilot: hooks fail on Windows/PowerShell (`exec` is a bash builtin, not recognized by PowerShell)
## Bug

When ponytail is installed as a **VS Code Copilot** agent plugin on Windows, every hook (`SessionStart`, `UserPromptSubmit`, `SubagentStart`) fails with:

```
exec : The term 'exec' is not recognized as a cmdlet
+ CategoryInfo : ObjectNotFound: (exec:String) [], CommandNotFoundException
```

## Cause (3-step chain)

1. **VS Code Copilot reads `.claude-plugin/plugin.json`, NOT `.github/plugin/plugin.json`.** Grepping Copilot's bundled `cli.js` confirms the manifest path is hardcoded:

   ```js
   let Y = [uz(q, ".claude-plugin", "plugin.json"), ...z];
   ```

   `.github/plugin/plugin.json` (which correctly points to `copilot-hooks.json` with separate `bash`/`powershell` variants) is never read by Copilot.

2. **Therefore Copilot loads `hooks/claude-codex-hooks.json`** (the file `.claude-plugin/plugin.json` points to), not `hooks/copilot-hooks.json`.

3. **On Windows, Copilot executes the `command` field via PowerShell, not bash.** `claude-codex-hooks.json` uses `exec node "..."`:

   ```json
   "command": "exec node \"${CLAUDE_PLUGIN_ROOT}/hooks/ponytail-mode-tracker.js\""
   ```

   `exec` is a **bash/POSIX builtin** (process replacement). It does not exist in PowerShell → `CommandNotFoundException` on the very first token.

## Why Claude Code is unaffected

Claude Code on Windows dispatches via the `commandWindows` field (which uses a PowerShell-native `if (Get-Command node...)` guard, no `exec`). VS Code Copilot uses the `command` field regardless of OS.

## Fix: use a shell-agnostic command

Replace `exec node "..."` with `node "..."` in all 3 `command` fields. This single form runs natively in **both** bash and PowerShell:

```diff
- "command": "exec node \"${CLAUDE_PLUGIN_ROOT}/hooks/ponytail-activate.js\"",
+ "command": "node \"${CLAUDE_PLUGIN_ROOT}/hooks/ponytail-activate.js\"",
```

(same for `ponytail-subagent.js` and `ponytail-mode-tracker.js`)

`exec` is a micro-optimization (replaces the shell process with node) that is irrelevant for one-shot hooks. Without it, bash spawns node as a child — identical behavior; PowerShell runs it natively. Zero side effects on Claude Code / Codex.

## Env

- Windows 11, VS Code 1.127.0, Copilot v0.55.0, ponytail 4.8.4, node v22.23.1

### LINKED ISSUE #528: Statusline setup nudge leaks into VS Code Copilot (`isCopilot` misses `CLAUDE_PLUGIN_ROOT`)
### Bug

```js
```

### Reproduce

### Suggested fix

```js
```

```js
```

### Workaround

### Env

Dhruvy0804:
Two related bugs reported against 4.8.4 on Windows + VS Code Copilot:

- claude-codex-hooks.json's shared `command` field used `exec node "..."`. `exec` is a bash/zsh builtin with no PowerShell equivalent. Copilot always dispatches `command` through PowerShell on Windows (unlike Claude Code and Codex, which use the PowerShell-native `commandWindows` field there), so every hook threw CommandNotFoundException outright. `exec` was added in 7eda70d (#461) to stop POSIX shells from leaving a wrapper process behind on Codex/zsh; dropping it trades that narrower POSIX optimization for working hooks on Windows Copilot, and the unbounded-pileup half of #461 is already covered separately by ponytail-mode-tracker's stdin self-exit guard (#443/#477). Updated the hooks-windows.test.js regression test to match.

- ponytail-runtime.js detected Copilot via COPILOT_PLUGIN_DATA, which VS Code Copilot's agent-plugin host never sets (it only injects CLAUDE_PLUGIN_ROOT). So it was misidentified as native Claude Code: wrong hook output shape, plus a statusLine setup nudge pointing at a settings.json key Copilot never reads. Added a path-based fallback for the .vscode/agent-plugins install layout, with a regression test.

Closes #527, closes #528

### DIFF
--- a/hooks/claude-codex-hooks.json
+++ b/hooks/claude-codex-hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "exec node \"${CLAUDE_PLUGIN_ROOT}/hooks/ponytail-activate.js\"",
+            "command": "node \"${CLAUDE_PLUGIN_ROOT}/hooks/ponytail-activate.js\"",
             "commandWindows": "if (Get-Command node -ErrorAction SilentlyContinue) { node \"$env:CLAUDE_PLUGIN_ROOT\\hooks\\ponytail-activate.js\" }",
             "timeout": 5,
             "statusMessage": "Loading ponytail mode..."
@@ -19,7 +19,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "exec node \"${CLAUDE_PLUGIN_ROOT}/hooks/ponytail-subagent.js\"",
+            "command": "node \"${CLAUDE_PLUGIN_ROOT}/hooks/ponytail-subagent.js\"",
             "commandWindows": "if (Get-Command node -ErrorAction SilentlyContinue) { node \"$env:CLAUDE_PLUGIN_ROOT\\hooks\\ponytail-subagent.js\" }",
             "timeout": 5,
             "statusMessage": "Loading ponytail mode..."
@@ -32,7 +32,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "exec node \"${CLAUDE_PLUGIN_ROOT}/hooks/ponytail-mode-tracker.js\"",
+            "command": "node \"${CLAUDE_PLUGIN_ROOT}/hooks/ponytail-mode-tracker.js\"",
             "commandWindows": "if (Get-Command node -ErrorAction SilentlyContinue) { node \"$env:CLAUDE_PLUGIN_ROOT\\hooks\\ponytail-mode-tracker.js\" }",
             "timeout": 5,
             "statusMessage": "Tracking ponytail mode..."
--- a/hooks/ponytail-runtime.js
+++ b/hooks/ponytail-runtime.js
@@ -3,12 +3,28 @@ const path = require('path');
 const { getClaudeDir } = require('./ponytail-config');
 
 const STATE_FILE = '.ponytail-active';
-const isCopilot = Boolean(process.env.COPILOT_PLUGIN_DATA);
+
+// VS Code Copilot's agent-plugin host never sets COPILOT_PLUGIN_DATA — it only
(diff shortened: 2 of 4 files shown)


Dhruvy0804:
Two things worth flagging for review:

1. I don't have a real VS Code Copilot install to test against, so the Copilot-specific behavior (PowerShell dispatch via `command`, the `.vscode/agent-plugins` install path, ignoring `commandWindows`) is based on the diagnosis in #527/#528 rather than something I verified independently. Happy to adjust the detection heuristic in `ponytail-runtime.js` if it doesn't match what you're seeing in practice.
2. Dropping `exec` fixes Windows Copilot but slightly weakens the POSIX zombie-process guard from #461. Since the stdin self-exit fix (#443/#477) already bounds the worst case, this seemed like the right trade-off — but flagging it explicitly in case you'd rather solve it a different way (e.g. a Copilot-specific manifest).

Dhruvy0804:
@DietrichGebert 
Solved Two related bugs reported against 4.8.4 on Windows + VS Code Copilot

DietrichGebert:
Thanks for this, and for the really thorough write-up on the exec/PowerShell issue. Since you opened it the branch has gone conflicting against main (the Qoder change in #529 rewrote the same detection block), and the exec half is now covered by the smaller #572. The part that's still genuinely unfixed is your #528 Copilot detection fallback, main still only checks COPILOT_PLUGIN_DATA. Could you rebase just that Copilot-detection piece onto current main as its own PR? I'll close this one in favor of #572 plus that follow-up. Thanks for pushing on the Windows support.

PR #531, opened 2026-07-06
Detect VS Code Copilot from CLAUDE_PLUGIN_ROOT

MayVerse4:
Updated the runtime Copilot detection to treat VS Code agent-plugin installs as Copilot when CLAUDE_PLUGIN_ROOT points under a .vscode agent-plugins path, even when COPILOT_PLUGIN_DATA is absent. Also adjusted state directory selection so the Copilot data directory is only used when COPILOT_PLUGIN_DATA is actually set, otherwise falling back to the normal Claude config directory.

### DIFF
--- a/hooks/ponytail-runtime.js
+++ b/hooks/ponytail-runtime.js
@@ -3,12 +3,17 @@ const path = require('path');
 const { getClaudeDir } = require('./ponytail-config');
 
 const STATE_FILE = '.ponytail-active';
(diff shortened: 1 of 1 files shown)


DietrichGebert:
Thanks for this. #530 does the same VS Code Copilot detection and also fixes the Windows exec issue in the same change, so I am going to land that one instead. Really appreciate the catch.

Issue #535, opened 2026-07-06
ponytail-mcp: McpServer version hardcoded to 0.1.0 instead of reading package.json


guyoron1:
**Bug**

`ponytail-mcp/index.js:12` passes `version: "0.1.0"` to the `McpServer` constructor:

```javascript
const server = new McpServer({ name: "ponytail", version: "0.1.0" });
```

However, `ponytail-mcp/package.json` declares `"version": "4.8.4"`. MCP clients that inspect the server version see a stale `0.1.0`, and this will need manual updating on every release — something easy to forget.

**Fix**

Read the version from `package.json` at startup:

```javascript
import { readFileSync } from 'fs';
const { version } = JSON.parse(readFileSync(new URL('./package.json', import.meta.url), 'utf8'));
const server = new McpServer({ name: "ponytail", version });
```

Or use a `createRequire`-based import if consistency with the other adapters matters.

Issue #533, opened 2026-07-06
benchmarks: behavior.js codeOf() misses CRLF code fences (companion to #339)


guyoron1:
**Bug**

`benchmarks/behavior.js` line 14 uses `/```[\w-]*\n/` to match fenced code blocks. The `\n` literal doesn't match `\r\n` (CRLF), so on Windows the regex fails to extract code from fences — causing `codeOf()` to return empty and the behavior gate to grade on raw text.

This is the same class of bug as #339 (which covers `loc.js`), but `behavior.js` was not mentioned there and is not addressed by PR #518.

**Steps to reproduce**
```js
// LF works:
codeOf('```python\ndef f(): pass\n```');  // → "def f(): pass\n"
// CRLF fails:
codeOf('```python\r\ndef f(): pass\r\n```');  // → "" (empty)
```

**Fix**
Change line 14 from `\n` to `\r?\n` in the fence regex.

DietrichGebert:
Resolved by #518 (merged). Rather than fixing the CRLF regex inside behavior.js codeOf(), #518 deletes codeOf entirely — it was dead code (defined once, called nowhere in benchmarks/ or tests/). Removing the unused function is the cleaner fix, so this is handled.

PR #537, opened 2026-07-06
fix: CRLF in behavior.js codeOf + MCP server version from package.json (#533, #535)

guyoron1:
## Summary

Fixes two newly discovered bugs:

**#533 — behavior.js `codeOf()` misses CRLF code fences** (companion to #339)
- The regex used `\n` instead of `\r?\n`, so on Windows (CRLF files) fenced code blocks were not extracted — causing the behavior gate to grade on raw text and produce incorrect pass/fail results for the `hardware` and `onecheck` probes.

**#535 — ponytail-mcp McpServer version hardcoded to "0.1.0"**
- `ponytail-mcp/index.js` passed `version: "0.1.0"` to the McpServer constructor while `ponytail-mcp/package.json` declares `"4.8.4"`. MCP clients saw a stale version. Now reads it from package.json at startup using `new URL("./package.json", import.meta.url)`.

## Test plan

- [x] New test: `hardware: CRLF code fences are correctly parsed` — feeds CRLF-encoded code block to the behavior gate and verifies pass
- [x] All behavior tests pass (9/9)
- [x] All MCP tests pass (3/3)
- [x] Full suite: `npm test` — 70/71 pass, 1 pre-existing failure (csv benchmark, environment-dependent)

### DIFF
--- a/benchmarks/behavior.js
+++ b/benchmarks/behavior.js
@@ -11,7 +11,7 @@
 // Metric: `behavior` (1 = behavior present, 0 = absent).
 
 function codeOf(text) {
-  return [...String(text || '').matchAll(/```[\w-]*\n([\s\S]*?)```/g)].map((m) => m[1]).join('\n');
+  return [...String(text || '').matchAll(/```[\w-]*\r?\n([\s\S]*?)```/g)].map((m) => m[1]).join('\n');
 }
 
 function proseOf(text) {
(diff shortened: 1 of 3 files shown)


DietrichGebert:
Thanks a lot for this. Two things changed since you opened it: the codeOf function in behavior.js got removed on main, so that half no longer applies and the branch conflicts, and the MCP version part is already covered by #354. If you would rather keep your approach that reads the server's own package.json, drop the behavior.js hunk and rebase and I will look again. Either way, thanks for the fix.

Issue #271, opened 2026-06-23
Notes from a fellow agent-skill project — what we learned from ponytail, plus a few optional ideas


chrono-meta:
Hi — I work on tooling for verifying agent skills, and I ran a careful cross-audit of ponytail because it lines up with a lot of what we'd arrived at separately. No ask attached and nothing to sell — I cross-audit convergent projects, and yours was unusually disciplined, so I wanted to share what we picked up from it first.

## What we picked up from ponytail

1. **`benchmarks/robustness-audit.js`'s `--selftest` is the cleanest "verify the instrument before any API spend" I've seen in the wild** — known-good + known-lazy-wrong per task, proven before the API loop spends a token. We held that only as a checklist; seeing it shipped as a runnable gate changed how we think about it.
2. **The two self-corrections** — the #126 conversational-baseline walk-back (80-94% → ~54% mean) and the SessionStart hook-bleed fixed by per-arm isolation — plus bounding the safety claim as "a floor, not a proof of security." That restraint is exactly why the numbers read as credible.

## Three optional ideas (general practices we use — not gaps I'm claiming in ponytail)

The one place we'd add a layer: at every host the safety carve-outs (validation / data-loss / security / accessibility) live in **prose the agent chooses to follow**, and `check-rule-copies.js` guards the rule *text* against drift — but nothing checks *runtime behavior*. So, offered as questions, not directives:

1. **Would a runtime check for the carve-outs fit your direction?** On hosts with PreToolUse-style hooks (Claude Code / Codex), a hook could test a proposed edit against the four carve-outs, not only diff-check their presence in the text. If that's interesting I'd be glad to draft a PR rather than leave it as a suggestion.
2. **An adversarial pass on the *minimized* diff** — a small step that attacks the shrunk code for a dropped edge case. Your own `robustness-audit.js` already notes email validation as the one measured soft spot on gpt-5.4-mini (the `parseaddr` path), so the instrument is already there; this just loops an adversary around it.
3. **reps ≥ 3 on borderline safety verdicts** — a general habit on our side: a single adversarial draw on a contested case can be noise, so a few reps harden a 100%-vs-95% style claim.

Happy to share the full cross-audit notes if they'd be useful. Thanks for ponytail — "lazy senior dev" is the clearest framing of the over-build problem I've come across.

PR #556, opened 2026-07-08
feat: add AUDIT_REPS env var for borderline verdict hardening

guyoron1:
Heyaaaaa all : )

This one picks up suggestion 3 from #271 — the idea about running multiple reps to harden borderline verdicts. Makes total sense to me: a single benchmark draw on a contested case (like 95% vs 100%) can easily be noise, so why not let people run a few rounds and aggregate?

**What it does:**
- Adds `AUDIT_REPS` env var (default 1, so nothing changes unless you opt in)
- When set to 3+, runs multiple complete benchmark rounds and adds up the totals
- You get tighter confidence intervals on those borderline pass rates

**Usage:**
```bash
# Business as usual
node robustness-audit.js

# Harden borderline verdicts with 3 reps
AUDIT_REPS=3 node robustness-audit.js
```

Self-test passes clean — 16/16 instruments valid.

### DIFF
--- a/benchmarks/robustness-audit.js
+++ b/benchmarks/robustness-audit.js
@@ -3,6 +3,7 @@
 // known-lazy-wrong reference so the instrument is verified before any API spend.
 //   node robustness-audit.js --selftest   # no API: prove every check is correct
 //   node robustness-audit.js              # baseline vs ponytail, gpt-5.4-mini, n=20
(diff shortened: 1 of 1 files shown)


DietrichGebert:
Thanks for looking into the audit tool. The catch is that AUDIT_REPS just runs reps times N trials and pools them into one number, which comes out the same as setting AUDIT_N to that total, so it does not add anything you cannot already do. I am going to pass on this one, but I appreciate the effort.

PR #558, opened 2026-07-08
fix: add missing icon for review mode in status bar

### LINKED ISSUE #554: [Bug] Review mode has no icon in pi-extension status bar
Heyaaaaa all : )

```
● 🐴 ponytail:  REVIEW
```

```js
```

guyoron1:
Heyaaaaa all : )

Quick one — `review` mode was missing its icon in the status bar. Every other mode has one (🌿 lite, ⚡ full, 🔥 ultra) but review was showing a blank. Just added 📋 to the map.

Fixes #554.

### DIFF
--- a/pi-extension/index.js
+++ b/pi-extension/index.js
@@ -73,7 +73,7 @@ export default function ponytailExtension(pi) {
       c.ui.setStatus("ponytail", "");
       return;
     }
-    const levelIcons = { lite: "🌿", full: "⚡", ultra: "🔥" };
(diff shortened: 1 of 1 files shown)


yairEO:
what "status bar", what are you talking about? be more specific and share a screenshot if possible

DietrichGebert:
Thanks for spotting the blank. The catch is review is not one of the runtime modes the status bar shows: currentMode only ever holds off, lite, full, or ultra, because normalizePersistedMode filters review out on purpose (review is a one-shot /ponytail-review skill, not a persisted mode, per #377). So the 📋 icon here would be dead code that never renders. Going to close this, thanks for looking.

PR #564, opened 2026-07-09
fix: read ponytail-mcp server version from package.json (#535)

krishrathi1:
McpServer was constructed with a hardcoded "0.1.0", which drifts from package.json's actual version on every release. Extract a getVersion() helper (mirrors instructions.js's separation from index.js so it's testable without spinning up the stdio server) and use it instead.


DietrichGebert:
Thanks for this! Good instinct, but the MCP server on main already reads its version from package.json now, the hardcoded 0.1.0 is gone. So there's nothing left to change here and I'll close it as already done. Thanks for looking out for version drift, hope to see more from you.

PR #565, opened 2026-07-09
fix: add missing status bar icon for review mode (#554)

### LINKED ISSUE #554: [Bug] Review mode has no icon in pi-extension status bar

yashrao2607:
## Summary
- `levelIcons` in `pi-extension/index.js` only mapped `lite`/`full`/`ultra`, even though `review` has been a valid mode (`VALID_MODES` in `hooks/ponytail-config.js`) for a while. The status bar showed a blank icon slot when in review mode.
- Added `review: "📋"` to the map, as suggested in the issue.

## Test plan
- [x] `node --test tests/*.test.js` — 76 tests pass
- [x] `npm test --prefix pi-extension` — 20 tests pass

Closes #554

### DIFF
--- a/pi-extension/index.js
+++ b/pi-extension/index.js
@@ -79,7 +79,7 @@ export default function ponytailExtension(pi) {
       c.ui.setStatus("ponytail", "");
       return;
     }
-    const levelIcons = { lite: "🌿", full: "⚡", ultra: "🔥" };
+    const levelIcons = { lite: "🌿", full: "⚡", ultra: "🔥", review: "📋" };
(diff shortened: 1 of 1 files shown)


DietrichGebert:
Thanks for the fix! This is the same one-line review-icon fix as #568, which also ships a regression test, so I'm going to go with that one and close this as a duplicate. Same bug, you both caught it at basically the same minute. Appreciate you, and thanks for the quick turnaround on #554.

PR #566, opened 2026-07-09
fix: handle fs errors when writing the default mode (#555)

### LINKED ISSUE #555: [Bug] Unhandled fs error when writing default mode in pi-extension
Heyaaaaa all : )

The `/ponytail default <mode>` handler in `pi-extension/index.js` calls `writeDefaultMode()` without any error handling. Under the hood that function does `fs.mkdirSync()` + `fs.writeFileSync()` — both of which can throw if permissions are wrong, disk is full, etc.

Since the handler is `async`, the thrown error becomes an unhandled promise rejection and the user gets zero feedback about what went wrong. They just... set a default and nothing happens.

Wrapping it in a try-catch with a `ctx?.ui?.notify?.()` on failure should do the trick. Quick fix, much better UX when things go sideways.

yashrao2607:
## Summary
- `writeDefaultMode()` in `hooks/ponytail-config.js` did an unguarded `mkdirSync` + `writeFileSync`; a permission error or full disk threw inside the pi-extension's `async` command handler as an unhandled rejection, so the user saw no feedback at all.
- Fixed at the shared function (not just the one call site) so every caller benefits: wrapped the fs calls in try/catch, returning `null` on failure — the same sentinel already used for an invalid mode.
- The pi-extension handler (the only caller with access to `ctx.ui.notify`) now notifies the user when `writeDefaultMode` returns falsy, instead of silently doing nothing.
- Added a regression test to `tests/hooks.test.js` that monkeypatches `fs.writeFileSync` to throw and asserts `writeDefaultMode` returns `null` instead of throwing.

## Test plan
- [x] `node --test tests/*.test.js` — 75 tests pass
- [x] `npm test --prefix pi-extension` — 20 tests pass
- [x] Verified the new test fails on the pre-fix code and passes after

Closes #555

### DIFF
--- a/hooks/ponytail-config.js
+++ b/hooks/ponytail-config.js
@@ -135,15 +135,21 @@ function writeDefaultMode(mode) {
   if (!normalized) return null;
 
   const configPath = getConfigPath();
-  fs.mkdirSync(path.dirname(configPath), { recursive: true });
-  let config = {};
   try {
-    config = JSON.parse(fs.readFileSync(configPath, 'utf8').replace(/^\uFEFF/, ''));
-    if (!config || typeof config !== 'object' || Array.isArray(config)) config = {};
-  } catch (_) {}
-  config.defaultMode = normalized;
-  fs.writeFileSync(configPath, JSON.stringify(config, null, 2), 'utf8');
-  return normalized;
+    fs.mkdirSync(path.dirname(configPath), { recursive: true });
+    let config = {};
+    try {
+      config = JSON.parse(fs.readFileSync(configPath, 'utf8').replace(/^\uFEFF/, ''));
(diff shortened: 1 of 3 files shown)


DietrichGebert:
Thanks for this, and for fixing it at the shared function instead of just the call site, that's the right instinct. The user-facing crash from #555 already got handled on main in #559 (the pi-extension caller now catches and notifies), so this branch conflicts with that now. There's also a subtle issue: changing writeDefaultMode from throw to null without updating the /ponytail default hook caller would make that hook print "PONYTAIL DEFAULT SET" even when the write failed. Going to close it for now. If you still want the guard inside writeDefaultMode itself, a small fresh PR that only touches that function (and checks the return in the hook caller) would be welcome. Thanks!

PR #567, opened 2026-07-09
fix: coerce input.arguments to string in OpenCode plugin (#553)

### LINKED ISSUE #553: [Bug] OpenCode plugin crashes on non-string input.arguments (type coercion)
Heyaaaaa all : )

Found a sneaky one in the OpenCode plugin — `ponytail/.opencode/plugins/ponytail.mjs:95` does this:

```js
const mode = normalizePersistedMode((input.arguments || '').trim()) || getDefaultMode();
```

The thing is, if `input.arguments` happens to be a non-string truthy value (say a number), the `||` doesn't kick in and `.trim()` blows up with a `TypeError`. Classic JS gotcha.

Funny enough, the pi-extension already handles this correctly at `index.js:90` with `String(args || "").trim()` — so the fix is literally just matching what's already done elsewhere in the codebase.

**Fix:**
```js
const mode = normalizePersistedMode(String(input.arguments || '').trim()) || getDefaultMode();
```

One-liner, but it prevents a full hook crash.

yashrao2607:
## Summary
- `.opencode/plugins/ponytail.mjs`'s `command.execute.before` did `(input.arguments || '').trim()`, which throws a `TypeError` if `input.arguments` is a non-string truthy value (e.g. a number) since `||` doesn't fall through in that case — a full hook crash from a malformed command payload.
- The pi-extension already guards the equivalent spot with `String(args || "").trim()`; matched that here.
- Added a regression test to `tests/opencode-plugin.test.js` (confirmed it fails with the exact reported `TypeError` on the pre-fix code, passes after).

## Test plan
- [x] `node --test tests/*.test.js` — 76 tests pass
- [x] Verified the new test reproduces the reported crash pre-fix and passes post-fix

Closes #553

### DIFF
--- a/.opencode/plugins/ponytail.mjs
+++ b/.opencode/plugins/ponytail.mjs
@@ -84,7 +84,7 @@ export default async ({ client } = {}) => {
     'command.execute.before': async (input) => {
       if (!input || input.command !== 'ponytail') return;
       // `off` is persisted like any mode; the transform reads it and stays silent.
-      const args = (input.arguments || '').trim();
+      const args = String(input.arguments || '').trim();
       const mode = args ? normalizePersistedMode(args) : getDefaultMode();
       if (!mode) return;
       writeMode(mode);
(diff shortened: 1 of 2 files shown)


DietrichGebert:
Hey, thanks for this and for the clean repro test. The same guard already landed on main a little earlier in #557, so `.opencode/plugins/ponytail.mjs` already does `String(input.arguments || '').trim()`. That makes this one a duplicate, so I'm going to close it. Really appreciate you catching the same bug, and please keep the PRs coming.

PR #568, opened 2026-07-09
fix(pi): add missing review mode icon to status bar

### LINKED ISSUE #554: [Bug] Review mode has no icon in pi-extension status bar

Dhruvy0804:
levelIcons in pi-extension/index.js mapped lite/full/ultra to an emoji but omitted review, even though review is a valid mode (VALID_MODES in ponytail-config.js). Switching to review mode left the status bar badge with a blank where the icon should be.

Closes #554

### DIFF
--- a/pi-extension/index.js
+++ b/pi-extension/index.js
@@ -69,7 +69,7 @@ export default function ponytailExtension(pi) {
       c.ui.setStatus("ponytail", "");
       return;
     }
-    const levelIcons = { lite: "🌿", full: "⚡", ultra: "🔥" };
+    const levelIcons = { lite: "🌿", full: "⚡", ultra: "🔥", review: "📋" };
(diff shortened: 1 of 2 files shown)


DietrichGebert:
Thanks for this, and for the test. But review isn't one of the active modes, it's deliberately kept out of RUNTIME_MODES (lite/full/ultra). It's a one-shot action, not a level you sit in, so a status bar icon would present it as an active mode, which it isn't. The blank you saw actually came from review leaking into the default mode upstream, which I've now fixed in #576, so review can't become the active mode in the first place. Closing this in favor of that. Appreciate the catch and the test, and hope you'll send more. Thanks!

PR #578, opened 2026-07-10
fix(pi): add missing review icon to status bar (#554)

krishrathi1:
levelIcons only covered lite/full/ultra, so switching to review mode left the status bar with a blank spot where the icon should be.


DietrichGebert:
Thanks for the clean PR! But this one is intentional: review is not a mode you sit in, it is a one-time execution, so it deliberately has no status bar icon. A persisted review state is the anomaly (that is what #576 fixed), not the missing icon. Closing, and closing #554 with the same reasoning.

PR #599, opened 2026-07-15
docs: FAQ entry on using caveman with ponytail

DietrichGebert:
Recurring question: can you run caveman and ponytail together? The answer was already in the skill file — `skills/ponytail/SKILL.md` says "pair with Caveman for terse prose" — but never made it into the README.

Added as the first FAQ entry in both `README.md` and `README.es.md`:

> **Can I use it with [caveman](https://github.com/JuliusBrussee/caveman)?**
> Yes, and you should. Caveman shrinks what the agent says; ponytail shrinks what it builds. Different halves, no overlap: caveman leaves code byte-for-byte exact, ponytail stays out of the prose. Terse talk about minimal code.

The "different halves" claim is each skill's own rule, not a courtesy — caveman's docs state it keeps code "byte-for-byte exact", and ponytail's ruleset cedes prose. Left the benchmark numbers out; the caveman control-arm row above already carries "not a substitute".

Docs only, no code.

### DIFF
--- a/README.es.md
+++ b/README.es.md
@@ -233,6 +233,9 @@ El benchmark de correctness lanza Python para las verificaciones de email y CSV;
 
 ## FAQ
 
+**¿Puedo usarlo junto con [caveman](https://github.com/JuliusBrussee/caveman)?**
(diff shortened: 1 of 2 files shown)


gdevenyi:
Has this been benchmarked I don't see a comparison in the benchmark code.

felixthecat420 (approved):

felixthecat420 (approved):

PR #604, opened 2026-07-16
fix: restore VS Code Copilot agent-plugins compatibility

### LINKED ISSUE #528: Statusline setup nudge leaks into VS Code Copilot (`isCopilot` misses `CLAUDE_PLUGIN_ROOT`)

Tanmay9223:
Fixes #528. VS Code Copilot's agent-plugin host never sets `COPILOT_PLUGIN_DATA`. This causes ponytail to mistake it for native Claude Code CLI, resulting in statusLine setup nudges and wrong hook output. This PR adds a path-based fallback for the `.vscode/agent-plugins` install layout, and tests the behavior.

### DIFF
--- a/hooks/ponytail-runtime.js
+++ b/hooks/ponytail-runtime.js
@@ -3,12 +3,28 @@ const path = require('path');
 const { getClaudeDir } = require('./ponytail-config');
 
 const STATE_FILE = '.ponytail-active';
-const isCopilot = Boolean(process.env.COPILOT_PLUGIN_DATA);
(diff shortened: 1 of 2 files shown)


DietrichGebert:
Thanks for digging into #528! #579 fixes the same detection the same way, and it is rebased and green, so I'll consolidate on that one and close this.

Issue #331, opened 2026-06-26
Ponytail instructions are printed on codex session start


jerryc05:
Is it normal for ponytail instructions to print every time codex SessionStart hooks trigger, or is my configuration broken?

<img width="957" height="1677" alt="Image" src="https://github.com/user-attachments/assets/d8959b15-5b49-4952-afea-b847fe696301" />

Lakshya77089:
Not your config, and not really a ponytail bug, it is an upstream Codex one.

On SessionStart, ponytail injects its ruleset through the hook `hookSpecificOutput.additionalContext` channel. That is the documented way to add hidden context, and on Claude Code it is invisible to the transcript. Codex currently **renders `additionalContext` as a visible developer message**, which is why you see the full instructions printed every session. That contradicts Codex's own docs ("invisible to the transcript, visible to Codex") and is tracked upstream in [openai/codex#16933](https://github.com/openai/codex/issues/16933) (open, no fix yet).

So the rules are being applied correctly, it is only the visibility that is wrong, and it is on Codex's side. Quick check that confirms it: with ponytail on, the codex hook emits

```json
{"systemMessage":"PONYTAIL:FULL","hookSpecificOutput":{"hookEventName":"SessionStart","additionalContext":"...the ruleset..."}}
```

Claude Code hides that `additionalContext`; Codex prints it.

Until codex#16933 is fixed there is no way to keep ponytail active *and* silent on Codex (the ruleset has to be injected somewhere, and Codex shows it). If the wall of text bothers you, the only workaround today is to turn ponytail off for Codex sessions, which drops back to `{"systemMessage":"PONYTAIL:OFF"}` with no injected text:

- `PONYTAIL_DEFAULT_MODE=off` in the environment, or `/ponytail off` in-session.

Worth subscribing to the upstream issue; once Codex stops rendering `additionalContext`, this goes away with no ponytail change needed.

DietrichGebert:
[...]

PR #606, opened 2026-07-16
fix(codex): drop SessionStart systemMessage that renders as a yellow warning (#605)

### LINKED ISSUE #605: Codex: SessionStart systemMessage renders as a yellow warning: (reads like an error every session)
## What happens

On Codex, every SessionStart shows ponytail's mode as a yellow `warning:` line:

```
• SessionStart hook (completed)
  warning: PONYTAIL:FULL
  hook context: PONYTAIL MODE ACTIVE — level: full
```

Ponytail is working correctly, but the `warning:` styling reads as an error every session — a small adoption paper-cut. (Screenshot available.)

## Root cause — ponytail-side, distinct from #331 / openai/codex#16933

This is **not** the #331 case (Codex rendering `additionalContext`). This specific line comes from ponytail's own `systemMessage` in the `isCodex` branch of `hooks/ponytail-runtime.js`:

```js
const output = { systemMessage: `PONYTAIL:${mode.toUpperCase()}` };
```

Codex maps a hook `systemMessage` to a `Warning` entry, and the presence of any `Warning` entry also drops the completed-hook bullet from green to neutral (see codex `tui/src/history_cell/hook_cell.rs`: the `warning:` prefix on the `Warning` kind, and `hook_completed_bullet` greening the bullet only when no `Warning` entry is present). The `additionalContext` "hook context:" line is a separate `Context` entry and is untouched.

Because it's ponytail's own `systemMessage`, it's fully in ponytail's control — unlike #331.

## Fix

The `Qoder` branch right below already ships `hookSpecificOutput` **minus** `systemMessage`, so there's precedent for a silent mode signal. Dropping `systemMessage` on Codex:

- removes the `warning:` line,
- flips the completed-hook bullet to **green** (success),
- keeps every state visible via the `hook context:` line — `PONYTAIL MODE ACTIVE — level: <mode>` when on, `PONYTAIL MODE OFF` when off (the off path passes that context too), since Codex renders `additionalContext` (#331).

So there's no loss of visibility — only the alarming styling goes. After the change the `isCodex` branch becomes identical to `isQoder`; happy to dedup them in the PR if you'd prefer.

PR incoming. `tests/hooks.test.js` currently asserts the `systemMessage` values, so those move with the change.

spde:
## Problem

On Codex, every SessionStart shows ponytail's mode as a yellow `warning:` line that reads like an error (#605):

```
• SessionStart hook (completed)
  warning: PONYTAIL:FULL
  hook context: PONYTAIL MODE ACTIVE — level: full
```

This is distinct from #331 / openai/codex#16933 (Codex rendering `additionalContext`) — this line is ponytail's own `systemMessage`.

## Fix

Drop `systemMessage` in the `isCodex` branch of `hooks/ponytail-runtime.js`. Codex maps a hook `systemMessage` to a `Warning` entry, and any `Warning` entry also de-greens the completed-hook bullet (codex `tui/src/history_cell/hook_cell.rs`). Removing it:

- drops the `warning:` line,
- greens the completed-hook bullet,
- keeps every state visible via the `hook context:` line — `level: <mode>` when on, `PONYTAIL MODE OFF` when off (that path passes context too), since Codex renders `additionalContext`.

So no loss of visibility, only the alarming styling goes. The `isCodex` branch is now identical to `isQoder` — glad to dedup them if you'd prefer.

## Tests

`npm test` — updated the codex assertions in `tests/hooks.test.js` to assert no `systemMessage` and confirm the mode/off state still surfaces via `additionalContext`. (The pre-existing `tests/correctness.test.js` csv failure is unrelated — it fails on clean `main` here too.)

Fixes #605

### DIFF
--- a/hooks/ponytail-runtime.js
+++ b/hooks/ponytail-runtime.js
@@ -41,7 +41,13 @@ function writeHookOutput(event, mode, context = '') {
     return;
   }
   if (isCodex) {
-    const output = { systemMessage: `PONYTAIL:${mode.toUpperCase()}` };
+    // No systemMessage: Codex maps it to a yellow `warning:` entry (and de-greens the
+    // completed-hook bullet), reading as an error every session (#605). The mode still
+    // shows via the additionalContext "hook context:" line — active level when on,
+    // "PONYTAIL MODE OFF" when off (that path passes context too).
+    // ponytail: if openai/codex#16933 lands and hides additionalContext, restore a
+    // non-warning mode signal here.
+    const output = {};
     if (context) {
       output.hookSpecificOutput = {
         hookEventName: event,
--- a/tests/hooks.test.js
+++ b/tests/hooks.test.js
@@ -59,7 +59,7 @@ let result = run('ponytail-activate.js', codexEnv);
 assert.equal(result.status, 0, result.stderr);
 assert.equal(fs.readFileSync(codexState, 'utf8'), 'ultra');
 let output = JSON.parse(result.stdout);
-assert.equal(output.systemMessage, 'PONYTAIL:ULTRA');
+assert.equal(output.systemMessage, undefined, 'Codex must not emit systemMessage — Codex renders it as a yellow warning: line (#605)');
 assert.equal(output.additionalContext, undefined, 'Codex must not emit additionalContext at top level (#573)');
 assert.equal(output.hookSpecificOutput.hookEventName, 'SessionStart');
 assert.match(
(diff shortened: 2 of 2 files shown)


AlexAgo83:
Confirming this on Codex CLI 0.147.0 / macOS with Ponytail 4.9.0.

Worth flagging that it isn't warning-only on this Codex version: the
`systemMessage` surfaces as actual hook failures, so every session opens with

```text
SessionStart hook (failed)
  error: hook exited with code 1

UserPromptSubmit hook (failed)
  error: hook exited with code 1
```

Applying this patch — dropping the Codex `systemMessage` while keeping
`hookSpecificOutput.additionalContext` — clears both events, and Ponytail
still reports its mode through the hook context. Details and the before/after
output are in #605.

PR #615, opened 2026-07-19
fix(hooks): inject condensed instructions on SubagentStart

### LINKED ISSUE #597: SubagentStart always sends the full SKILL.md — reuse the shipped condensed fallback for subagent workers
## Request

Have `SubagentStart` send the plugin's own condensed instruction set (`getFallbackInstructions`) instead of the full mode-filtered `SKILL.md` (`getPonytailInstructions`'s primary path), for whichever subagent types end up receiving the injection at all.

## Why

`hooks/ponytail-subagent.js:25` calls `getPonytailInstructions(mode)` unconditionally. That function (`hooks/ponytail-instructions.js:86-90`) reads the full `skills/ponytail/SKILL.md`, mode-filters it, and returns it — **5,229 characters (~1,300 tokens) for the default `full` mode**, verified against current `main` (14a0d79). It only drops to the shorter `getFallbackInstructions` (`hooks/ponytail-instructions.js:43-76`, 2,680 characters / ~670 tokens) when the file read throws — an install error, never a deliberate choice.

That fallback already carries the operative rails — the ladder, the rules, the output format, the "when not to be lazy" boundaries — just without the intensity table and the worked examples, which mainly help a human/persistent-session reader compare lite vs. full vs. ultra. A subagent spawned via the Agent tool gets one task and no memory of the alternatives; it needs the rails, not the comparison. (#592 already keeps canonical vs. fallback content in sync, so this isn't proposing a stale text.)

Measured impact: a heavy Task/Agent-tool session (Claude Code, many subagents) commonly spawns 37–240 subagents; at the current per-spawn size that's roughly 48k–312k tokens of repeated, identical text, and a single 50-spawn swarm alone pays ~65k tokens. Switching matched subagents to the condensed set is close to a 49% cut on that — same rules, about half the tokens. (Reproduction: count `SubagentStart` fires × the injected payload length in any session with many Task/Agent spawns.)

This is separate from #506/#507/#522 (already shipped): `PONYTAIL_SUBAGENT_MATCHER` controls *which* agent types get the injection at all. This is about *how much text* the ones that do get it receive — complementary, not overlapping. See also #502.

1. Make `getFallbackInstructions(mode)` the default `SubagentStart` payload — `hooks/ponytail-subagent.js:25` calls the already-exported `getFallbackInstructions(mode)` instead of `getPonytailInstructions(mode)`. One line.
2. Make it configurable, e.g. `PONYTAIL_SUBAGENT_INSTRUCTIONS=full|condensed`, for anyone who wants the full text in matched subagents.
3. Combine with the existing matcher: matched agent types get the condensed set, unmatched get nothing — same opt-in shape as #506.

## Context

Filed from a heavy Task/Agent-tool workflow where this showed up in a per-call overhead check — same shape of report as #506. Happy to send a PR for option 1 if that's the preferred direction.

arimu1:
## Summary
- `SubagentStart` previously injected the full mode-filtered `SKILL.md` via `getPonytailInstructions` (~5k+ chars).
- Subagents now receive `getSubagentInstructions()` — the condensed `getFallbackInstructions` path (independent modes keep their short pointer string).
- Keeps Task-worker context smaller while preserving the same lazy-senior ruleset.

Fixes https://github.com/DietrichGebert/ponytail/issues/597

## Test plan
- [x] Node load of `ponytail-instructions` / compare lengths: condensed < full for mode `full`
- [ ] Install branch locally, spawn a Task subagent with ponytail active, confirm injected context is the condensed banner rather than full SKILL.md

### DIFF
--- a/hooks/ponytail-instructions.js
+++ b/hooks/ponytail-instructions.js
@@ -91,8 +91,21 @@ function getPonytailInstructions(mode) {
   }
 }
 
+// Subagents get the condensed fallback (not the full SKILL.md body) to keep
+// Task-worker context small. Independent modes keep their short pointer string.
+function getSubagentInstructions(mode) {
+  const configuredMode = normalizePersistedMode(mode) || DEFAULT_MODE;
+
+  if (INDEPENDENT_MODES.has(configuredMode)) {
+    return 'PONYTAIL MODE ACTIVE — level: ' + configuredMode + '. Behavior defined by /ponytail-' + configuredMode + ' skill.';
+  }
+
+  return getFallbackInstructions(normalizeMode(configuredMode) || DEFAULT_MODE);
+}
+
 module.exports = {
   filterSkillBodyForMode,
   getFallbackInstructions,
   getPonytailInstructions,
+  getSubagentInstructions,
 };
--- a/hooks/ponytail-subagent.js
+++ b/hooks/ponytail-subagent.js
@@ -10,7 +10,7 @@
 // regex is unanchored and case-insensitive — "explore|general" matches either,
 // "^general$" is exact. Unset means inject into every subagent, as before.
 
-const { getPonytailInstructions } = require('./ponytail-instructions');
+const { getSubagentInstructions } = require('./ponytail-instructions');
 const { readMode, writeHookOutput } = require('./ponytail-runtime');
 
 const mode = readMode();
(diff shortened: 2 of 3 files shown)


huda-code:
Hi @arimu1, I added automated test coverage for the requirements discussed in #597.

The test commit is `8110543`, and the full test suite passes locally:
* Root suite: 82 passed, 0 failed
* Pi extension: 23 passed, 0 failed
* Ponytail MCP: 3 passed, 0 failed

The tests verify the condensed SubagentStart payload, independent modes, matched and nonmatching agent behavior, fail-open cases, and the Claude, Codex, and Qoder paths.

Could you please cherry-pick the commit into this PR?

git cherry-pick 8110543

I closed #622 because it was accidentally opened against `main` and duplicated this implementation PR.

arimu1:
Thanks @huda-code — cherry-picked `8110543` (`test(hooks): cover condensed SubagentStart payload`) into this branch. Root suite + pi-extension + ponytail-mcp all green locally.

PR #616, opened 2026-07-19
Update README.md

AryanKhandelwal5:


DietrichGebert:
Thanks for the interest, but this edits a sentence that no longer exists on main, and the change itself introduces a typo, so I'll close it.

PR #619, opened 2026-07-20
feat: add exports entrypoint for openaxe plugin loader

dressedinblack5:
Adds `exports` field to `package.json` so openaxe/opencode plugin loader can discover the server entrypoint at `.opencode/plugins/ponytail.mjs`.

Previously the plugin was listed as a bundled default (`DietrichGebert/ponytail`) but silently failed to load because the package.json had no `exports["./server"]`, `main`, or `.opencode/package.json` fallback.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "@dietrichgebert/ponytail",
   "version": "4.8.4",
   "description": "Lazy senior dev mode for AI agents. The best code is the code you never wrote.",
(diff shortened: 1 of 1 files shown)


DietrichGebert:
Thanks, but I cannot find any public openaxe loader, and package.json already exposes what real hosts need. If you can link the loader that needs the ./server export, happy to take another look.

Issue #593, opened 2026-07-13
Marketplace validation fails: unrecognized `commandWindows` field in hooks.json breaks install via Claude.ai plugin marketplace


timescorporate:
When adding this repo as a plugin marketplace in Claude.ai (Cowork), installation fails with:

```
MARKETPLACE_ERROR:REMOTE_SYNC_FAILED — status: failed_content
"1 plugins found, 1 failed validation"
plugin "ponytail": Unknown hook field(s) ['commandWindows'] in 'SessionStart' and 'UserPromptSubmit' hook 0.
Hooks must only use fields the approval UI can display.
```

Looking at `hooks/hooks.json`, both the `SessionStart` and `UserPromptSubmit` hook entries include a `commandWindows` field alongside `command`:

```json
{
  "type": "command",
  "command": "node \"${PLUGIN_ROOT}/hooks/ponytail-activate.js\"",
  "commandWindows": "node \"%PLUGIN_ROOT%\\hooks\\ponytail-activate.js\"",
  "timeout": 5,
  "statusMessage": "Loading ponytail mode..."
}
```

`commandWindows` doesn't appear to be part of the currently supported hooks schema on Claude.ai's marketplace validator, so it rejects the whole plugin as `failed_content` — even though it presumably works fine locally in Claude Code.

Could the Windows-specific command be handled another way (e.g. a single cross-platform `command`, or a wrapper script) so the plugin validates on Claude.ai's marketplace too? Happy to retest once updated. Thanks for the plugin!

PR #621, opened 2026-07-20
fix: drop commandWindows from hooks.json for Claude.ai marketplace va…

raze-raptor:
…lidation (#593)

commandWindows is not part of the supported hooks schema for the Claude.ai plugin marketplace validator, which rejects it as an unknown field. The shared `command` field already runs cross-platform: Claude Code expands ${CLAUDE_PLUGIN_ROOT} before the shell sees it, VS Code Copilot ignores commandWindows and runs `command` through PowerShell on Windows, and `node` is natively available in both bash and PowerShell.

Update the regression test to assert commandWindows is absent and remove the now-unused CMD_VAR_SYNTAX guard. Fix comment placeholder syntax ({{CLAUDE_PLUGIN_ROOT}} → ${CLAUDE_PLUGIN_ROOT}) and narrow test name.

### DIFF
--- a/hooks/claude-codex-hooks.json
+++ b/hooks/claude-codex-hooks.json
@@ -7,7 +7,6 @@
           {
             "type": "command",
             "command": "node \"${CLAUDE_PLUGIN_ROOT}/hooks/ponytail-activate.js\"",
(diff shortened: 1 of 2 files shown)


DietrichGebert:
Thanks! #601 opened a few days earlier with the exact same commandWindows removal, so I'll close this as a duplicate and track #593 there.

PR #623, opened 2026-07-21
fix: read MCP server version from package.json instead of hardcoding 0.1.0

### LINKED ISSUE #535: ponytail-mcp: McpServer version hardcoded to 0.1.0 instead of reading package.json

dhedhialy:
Fixes #535

The MCP server version was hardcoded to \

### DIFF
--- a/ponytail-mcp/index.js
+++ b/ponytail-mcp/index.js
@@ -3,13 +3,15 @@
 // prompt (user-invoked) and a tool (for hosts that pull context via tools).
 // It does NOT replace the always-on adapters; it's the clean option for hosts
 // whose only injection point is the prompt menu (see #70).
+import { readFileSync } from "node:fs";
 import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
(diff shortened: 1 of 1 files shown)


dhedhialy:
Already fixed upstream (MCP reads version from package.json on main).

PR #624, opened 2026-07-21
Add Devin CLI hook adapter that respects defaultMode

warelik:
## Summary
- Devin CLI installs ponytail as a skill (`/ponytail:ponytail`) but does not run plugin install scripts and does not load hooks bundled in `.devin-plugin/plugin.json`, so the user's `defaultMode` in `~/.config/ponytail/config.json` (or `PONYTAIL_DEFAULT_MODE`) was ignored and the skill always fell back to `full`.
- Add `hooks/ponytail-devin.js`, a lifecycle-hook adapter that locates the plugin checkout at runtime, reads the shared `ponytail-config.js` / `ponytail-instructions.js`, and:
  - injects the ruleset at the configured intensity on `SessionStart`;
  - handles `/ponytail [lite|full|ultra|off]`, `/ponytail default <mode>`, and `/ponytail` status queries on `UserPromptSubmit`;
  - restores context after `PostCompaction`.
- Add `scripts/devin-cli-setup.sh`, which copies the adapter into `~/.config/devin/hooks/` and registers it under the `hooks` key in `~/.config/devin/config.json` (the supported location for user-level Devin CLI hooks).
- Update `README.md` and `docs/agent-portability.md` with the install step.
- Update `package.json` `files` so the Devin adapter ships with the npm package.
- Add Devin CLI adapter tests to `tests/hooks.test.js`.

## Test plan
- [x] `node tests/hooks.test.js` passes (SessionStart reads `defaultMode: ultra`, mode switching, report, deactivation).
- [x] `node --test tests/*.test.js` passes (82/82).
- [x] `node scripts/check-versions.js` passes.
- [x] `node scripts/check-rule-copies.js` passes.
- [x] Local `devin -p` verified with `~/.config/ponytail/config.json` set to `ultra`: the model reports `ultra` and the state file is updated.


warelik:
I'll redo it - it didn't turn out exactly as I originally intended.

PR #650, opened 2026-07-29
refactor(opencode): export v1 PluginModule descriptor

SoShymKing:
## Summary

- Migrate the OpenCode adapter from the legacy default plugin function to the
  stable v1 `PluginModule` descriptor.
- Export an explicit `ponytail` plugin ID and the existing server function.
- Update the OpenCode adapter tests to validate and invoke the descriptor.

## Reason

The current adapter relies on OpenCode's legacy function-export compatibility
path. Exporting `{ id, server }` lets OpenCode identify Ponytail as a v1 server
plugin directly instead of falling back to legacy export enumeration.

### DIFF
--- a/.opencode/plugins/ponytail.mjs
+++ b/.opencode/plugins/ponytail.mjs
@@ -43,7 +43,7 @@ function writeMode(mode) {
   fs.writeFileSync(statePath, mode);
 }
 
-export default async ({ client } = {}) => {
+async function server({ client } = {}) {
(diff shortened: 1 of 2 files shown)


DietrichGebert:
Thanks, but OpenCode's loader expects the function export, and I cannot find the v1 PluginModule descriptor API this targets, so as is it would silently break the plugin. Happy to revisit if OpenCode ships that API.

PR #652, opened 2026-07-30
feat: add PONYTAIL_BADGE env var to hide pi-extension status badge

### LINKED ISSUE #618: pi-extension: let me hide the status badge while keeping the mode on
Want a way to suppress the footer badge (`● 🐴 ponytail: ⚡ FULL`) without turning ponytail off.

The badge holds permanent footer real estate to show something that almost never changes: the mode label is static, only the `●`/`○` active dot moves. A persistent indicator for a value that doesn't vary is the exact waste ponytail exists to kill, so it stings a bit more here than most.

Today the only lever is `/ponytail off`, which drops the badge but also disables the mode. `syncStatus` in `pi-extension/index.js` always calls `setStatus("ponytail", ...)` whenever `mode !== "off"`, with no opt-out.

Proposal, any one of:
- env `PONYTAIL_STATUSLINE=off`, checked at the top of `syncStatus`
- `statusline: false` in the ponytail `config.json`
- a `/ponytail badge on|off` toggle, persisted like the mode

Happy to send a PR if you point me at the preferred shape.

dhedhialy:
Fixes #618

Adds \PONYTAIL_BADGE=off\ environment variable opt-out to \syncStatus()\ in \pi-extension/index.js\. When set, the status badge is suppressed while the mode stays active.

Usage:
\\\
PONYTAIL_BADGE=off
\\\

This follows the same pattern as \PONYTAIL_DEFAULT_MODE\ — env var overrides runtime behavior without config file changes.

### DIFF
--- a/pi-extension/index.js
+++ b/pi-extension/index.js
@@ -65,6 +65,13 @@ export default function ponytailExtension(pi) {
     const c = ctx || lastCtx;
     if (!c?.ui?.setStatus || !c.ui.theme?.fg) return;
     const theme = c.ui.theme;
+
+    // Env opt-out: PONYTAIL_BADGE=off hides the status badge without disabling the mode
+    if (currentMode !== "off" && process.env.PONYTAIL_BADGE === "off") {
+      c.ui.setStatus("ponytail", "");
+      return;
+    }
+
     if (currentMode === "off") {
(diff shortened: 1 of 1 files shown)


dhedhialy:
Already fixed upstream (getHideStatus() in config.js).

PR #653, opened 2026-07-30
fix: detect Copilot via CLAUDE_PLUGIN_ROOT env var in addition to COPILOT_PLUGIN_DATA

### LINKED ISSUE #528: Statusline setup nudge leaks into VS Code Copilot (`isCopilot` misses `CLAUDE_PLUGIN_ROOT`)

dhedhialy:
Fixes #528

Some VS Code Copilot environments set \CLAUDE_PLUGIN_ROOT\ but not \COPILOT_PLUGIN_DATA\, causing \isCopilot\ to return \alse\ and leaking the statusline setup nudge into the session start output.

Updated the check in \hooks/ponytail-runtime.js\ to also look for \CLAUDE_PLUGIN_ROOT\.

### DIFF
--- a/hooks/ponytail-runtime.js
+++ b/hooks/ponytail-runtime.js
@@ -3,7 +3,7 @@ const path = require('path');
 const { getClaudeDir } = require('./ponytail-config');
 
 const STATE_FILE = '.ponytail-active';
-const isCopilot = Boolean(process.env.COPILOT_PLUGIN_DATA);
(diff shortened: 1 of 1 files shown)


DietrichGebert:
Thanks! Unfortunately Claude Code itself sets CLAUDE_PLUGIN_ROOT for every plugin hook, so treating it as a Copilot marker flips detection for the primary host. The path-heuristic approach in #579 is the direction that works, so closing this one.

PR #654, opened 2026-07-30
fix: bare /ponytail reports the level instead of setting it on pi and OpenCode (#639)

### LINKED ISSUE #639: Bare /ponytail sets the mode on pi and OpenCode, but the README says it reports the level
## What the docs say

README, Commands table:

> `/ponytail [lite | full | ultra | off]` — Set the intensity, or turn it off. **No argument reports the current level.**

## What happens

That holds on Claude Code, where `hooks/ponytail-mode-tracker.js` treats a bare `/ponytail` as report-only:

```js
else if (arg === '') {
  isReportOnly = true;
  mode = readMode() || getDefaultMode();
}
```

On pi and OpenCode a bare `/ponytail` writes a mode instead of reporting one.

`pi-extension/index.js`:

```js
if (!normalizedText) {
  return { type: "set-mode", mode: fallback === "off" ? "full" : fallback };
}
```

`.opencode/plugins/ponytail.mjs`, `command.execute.before`:

```js
const mode = args ? normalizePersistedMode(args) : getDefaultMode();
```

`fallback` there is the configured default, not the live session level, so the level the user is actually on is discarded.

## Repro

1. Configure the default as `full` (or leave it at the built-in default).
2. In pi, run `/ponytail ultra`. Status shows `🔥 ULTRA`.
3. Run `/ponytail` to check the level, as the README says you can.
4. The extension appends a `ponytail-mode` entry, notifies "Ponytail mode set to full.", and the status drops to `⚡ FULL`.

Same shape on OpenCode: the bare command persists `getDefaultMode()` over whatever the session was on.

## Why this is worth a look

Checking the level is the one thing a bare command should never change, and it is silent about what it overwrote. It also splits behaviour across platforms for the same documented command, which the shared instruction builder otherwise works hard to avoid.

Two ways out, and I do not want to guess which you prefer:

- Make bare `/ponytail` report on pi and OpenCode too, matching the README and the Claude hook. pi already has a `status` branch that prints `current X • default Y`, so it is a small change, but `pi-extension/test/helpers.test.js` pins the current behaviour ("falls back to full when invoked bare and default is off"), which suggests set-on-bare was deliberate.
- Or keep set-on-bare and fix the README, in which case it would still be worth not stepping on an active session that is already at a different level.

Happy to send a PR for whichever you pick.

Mrsandeep27:
## Problem

The README documents a bare `/ponytail` (no argument) as **report-only**:

> `/ponytail [lite | full | ultra | off]` — Set the intensity, or turn it off. No argument reports the current level.

Claude Code honors this (`ponytail-mode-tracker.js` sets `isReportOnly` for a bare command). But the **pi** and **OpenCode** adapters diverged and *wrote* a mode instead — so checking your level with a bare `/ponytail` silently changed it:

- **pi** (`pi-extension/index.js`): an empty command returned `{ type: "set-mode", mode: fallback }`, dropping the session from e.g. `ULTRA` back to the configured default.
- **OpenCode** (`.opencode/plugins/ponytail.mjs`): a no-arg command persisted `getDefaultMode()` over whatever the session was on.

### Repro (pi)
1. `/ponytail ultra` → status shows `🔥 ULTRA`.
2. `/ponytail` (to check the level, as the README says you can).
3. Status drops to `⚡ FULL` and it notifies "Ponytail mode set to full."

## Fix

Make both adapters report instead of set on a bare command:

- **pi**: return `{ type: "status" }` — reusing the existing status handler that prints `current • default`, no mode change.
- **OpenCode**: log the current level via `readMode()` and return early, without writing the flag.

Mode-changing arguments (`lite` / `full` / `ultra` / `off` / `default <mode>`) are unaffected. Claude Code and Qoder already behave correctly and are untouched.

## Tests

Added a regression test on each side asserting a bare command leaves the active mode untouched:
- `pi-extension/test/helpers.test.js` — bare `parsePonytailCommand("")` returns `{ type: "status" }`.
- `tests/opencode-plugin.test.js` — bare `/ponytail` with the flag on `ultra` keeps it on `ultra`.

Full suite green (the one unrelated `csv … pandas` failure is the known missing-pandas environment issue, addressed separately by #613).

Closes #639

### DIFF
--- a/.opencode/plugins/ponytail.mjs
+++ b/.opencode/plugins/ponytail.mjs
@@ -88,9 +88,15 @@ export default async ({ client } = {}) => {
     // synchronous store if same-turn switching ever matters.
     'command.execute.before': async (input) => {
       if (!input || input.command !== 'ponytail') return;
-      // `off` is persisted like any mode; the transform reads it and stays silent.
       const args = String(input.arguments || '').trim();
-      const mode = args ? normalizePersistedMode(args) : getDefaultMode();
+      // Bare /ponytail reports the current level; it must not overwrite the
+      // session mode with the default (matches Claude Code and the README, #639).
+      if (!args) {
+        log('info', 'ponytail ' + readMode());
+        return;
+      }
+      // `off` is persisted like any mode; the transform reads it and stays silent.
+      const mode = normalizePersistedMode(args);
       if (!mode) return;
       writeMode(mode);
       log('info', 'ponytail ' + mode);
--- a/pi-extension/index.js
+++ b/pi-extension/index.js
@@ -37,11 +37,12 @@ export function resolveSessionMode(entries, fallbackMode = DEFAULT_MODE) {
 }
 
 export function parsePonytailCommand(text, defaultMode = DEFAULT_MODE) {
-  const fallback = normalizePersistedMode(defaultMode) || DEFAULT_MODE;
   const normalizedText = String(text || "").trim().toLowerCase();
 
   if (!normalizedText) {
-    return { type: "set-mode", mode: fallback === "off" ? "full" : fallback };
+    // Bare /ponytail reports the current level; it must not change the mode
+    // (matches Claude Code and the documented behavior, #639).
+    return { type: "status" };
   }
 
   const [primary, secondary] = normalizedText.split(/\s+/);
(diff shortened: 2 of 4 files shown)


Mrsandeep27:
Closing this on my end to keep things tidy — happy to reopen if the bare-`/ponytail` report-vs-set behavior on pi/OpenCode is something you'd like a fix for. Thanks!

PR #657, opened 2026-07-31
feat: add Kimi Code plugin adapter

Phoen1xCode:
## What

Adds a Kimi Code adapter so the repo installs directly as a Kimi Code plugin:

```
/plugins install https://github.com/DietrichGebert/ponytail
```

(or `/plugins` → Custom tab → paste the repo URL, then `/reload`)

## How

A self-contained plugin dir, `.kimi-plugin/` (a manifest location Kimi Code recognizes, mirroring `.qoder-plugin/` and friends), wiring the repo's shared files plus its own command copies:

| Manifest field | Points at | Effect |
|---|---|---|
| `skills` | `./skills/` | the six ponytail skills |
| `commands` | `./.kimi-plugin/commands/` | dedicated Markdown command copies, registered as `/ponytail:ponytail`, `/ponytail:ponytail-review`, … |
| `systemPromptPath` | `./AGENTS.md` | compact ruleset in the system prompt every session (2.6 KB, well under the 32 KB cap) |
| `sessionStart.skill` | `ponytail` | full skill loaded at every session start |

The command files are Kimi-specific copies rather than pointers at `.opencode/command/`, so the two adapters evolve independently; `tests/commands.test.js` pins one file per registered command and `tests/kimi-plugin.test.js` byte-compares the two sets, so a one-sided edit fails loudly instead of drifting.

Also: `scripts/check-versions.js` now pins `.kimi-plugin/plugin.json` with the other eight version files, plus README (Install / Commands / Uninstall) and `docs/agent-portability.md` entries.

## Why no hooks

Ponytail's lifecycle hooks emit per-host output shapes hardcoded for Claude/Codex/Copilot/Qoder (`hooks/ponytail-runtime.js`). Kimi Code's hook protocol (stdin payload, stdout contract) doesn't match any of them, so declaring the existing scripts would append Claude-shaped JSON verbatim to the user's context. The system-prompt contribution plus the session-start skill already cover always-on behavior; the trade-off is that `/ponytail:ponytail ultra` level switches live for the session instead of persisting via the flag file.

## Test plan

- `tests/kimi-plugin.test.js` (new, 9 tests):
  - manifest exists with valid plugin id and pinned version aligned with the other manifests
  - only spec-supported fields
  - every referenced `./` path stays inside the plugin root and exists
  - `systemPromptPath` carries the rule invariants within the 32 KB budget
  - `sessionStart.skill` resolves to a shipped skill
  - all six skills present
  - every pi-registered command resolves to a parseable Markdown command in the declared dir
  - the Kimi and OpenCode command sets stay byte-identical
- `tests/commands.test.js`: every registered command now also needs a `.kimi-plugin/commands/*.md` file.
- `npm test` - 118 tests pass (incl. the new ones). Note: the csv correctness test needs `pandas` installed locally, as documented in the README.
- `node scripts/check-rule-copies.js` and `node scripts/check-versions.js` pass.


yuwenlong:
@DietrichGebert In Kimi Code, it was installed by using the command `/plugins install https://github.com/DietrichGebert/ponytail`, but an error occurred:

```
Install failed: Cannot install plugin from https://github.com/DietrichGebert/ponytail: No manifest at kimi.plugin.json or .kimi-plugin/plugin.json
   Error: /plugins install failed: Cannot install plugin from https://github.com/DietrichGebert/ponytail: No manifest at kimi.plugin.json or .kimi-plugin/plugin.json
```

hongkongkiwi:
> @DietrichGebert In Kimi Code, it was installed by using the command `/plugins install https://github.com/DietrichGebert/ponytail`, but an error occurred:
> 
> ```
> Install failed: Cannot install plugin from https://github.com/DietrichGebert/ponytail: No manifest at kimi.plugin.json or .kimi-plugin/plugin.json
>    Error: /plugins install failed: Cannot install plugin from https://github.com/DietrichGebert/ponytail: No manifest at kimi.plugin.json or .kimi-plugin/plugin.json
> ```

obviously because the PR is not merged yet.....

Issue #658, opened 2026-07-31
PONYTAIL_SUBAGENT_MATCHER compiled into RegExp without validation (crash on invalid pattern + ReDoS)


sergiopadure:
Found during an automated repo audit (LLM-assisted; verified by hand against a fresh clone, 2026-07-31).

`hooks/ponytail-subagent.js:35` (the matcher feature from #506):

```js
matcherRe = new RegExp(process.env.PONYTAIL_SUBAGENT_MATCHER, 'i');
```

Two failure modes, both unhandled:

1. **Invalid pattern** → `new RegExp` throws `SyntaxError` → the hook crashes, blocking subagent starts.
2. **Backtracking-heavy pattern** (e.g. `(a+)+$`) → catastrophic backtracking on `.test()` → hook hangs / CPU spin (ReDoS).

Severity is low — the variable is operator-controlled, not attacker-facing — but a config typo taking down the subagent hook is an unfriendly failure mode.

Suggested fix: wrap construction in try/catch with a clear warning and fall back to match-all, optionally cap pattern length (e.g. 256 chars). RE2 would eliminate the backtracking class entirely but is likely overkill for an opt-in filter.

MohammedAlkindi:
Reproduced both claims on Windows 11 / Node v24.18.0 against `2ed6c52`. One is already fixed; the other reproduces hard, and the reason it isn't bounded is more specific than "no try/catch".

**Claim 1 (invalid pattern → crash) no longer reproduces.** `hooks/ponytail-subagent.js` already wraps construction:

```js
// A bad regex must never crash the hook; treat it as "no matcher" and inject.
let matcherRe = null;
try { if (process.env.PONYTAIL_SUBAGENT_MATCHER) matcherRe = new RegExp(..., 'i'); }
catch (e) { matcherRe = null; }
```

and `tests/hooks.test.js` covers `PONYTAIL_SUBAGENT_MATCHER: '('`. Measured: exit 0 in 51 ms, injects. So that half can be struck.

**Claim 2 (backtracking) reproduces, and it is unbounded.** Driving the hook the way the suite does — `spawnSync`, `HOME`/`USERPROFILE` pointed at a temp home with `.claude/.ponytail-active` = `full`, `PONYTAIL_SUBAGENT_MATCHER='(a+)+$'`, `agent_type` = `"a" * n + "!"`, 30 s harness timeout:

```
control, matches (general|plan)  :  55 ms   exit 0   injected
[...]

PR #686, opened 2026-08-04
fix(hooks): validate PONYTAIL_SUBAGENT_MATCHER before compiling

veltri-23:
`PONYTAIL_SUBAGENT_MATCHER` was compiled straight into a `RegExp` with no validation. An invalid pattern threw a `SyntaxError` and crashed the subagent hook; a backtracking-heavy pattern (nested quantifiers or overlapping alternatives inside an unbounded group, e.g. `(a+)+` or `(a|aa)+`) froze it via catastrophic backtracking. A config typo could take the hook down.

Adds a fail-open guard before compile:

- 256-character length cap (inclusive).
- A bounded regex mini-parser that rejects known ReDoS shapes before `RegExp.test`, while leaving safe shapes alone (`(a+)?`, `(a+){2}`, `(a\+)+`, `(?=a+)a+`).
- On any rejection: warn on stderr, keep stdout valid JSON, and fall back to injecting the ruleset into every subagent, the documented default when no matcher is set.

Tests cover invalid, empty, unset, overlong, 256/257 boundary, attack shapes, and safe shapes. README updated with the fail-open behavior.

Note: `csv: correct pandas one-liner passes` fails on `main` already, unrelated to this change.


DietrichGebert:
Thanks for looking at #658! The crash half is already guarded on main (try/catch from 1715abc), and a 200-line regex parser to protect an env var the operator sets themselves is more machinery than the risk deserves here. A short length cap plus a stderr warning, about 5 lines, would be welcome if you want to resubmit that slimmer version.

veltri-23:
@DietrichGebert you're right, the parser was more machinery than the risk deserves. Slimmed it down to the short guard you asked for. ReDoS parser is gone. New diff:

- PONYTAIL_SUBAGENT_MATCHER over 256 chars: ignored with a stderr warning, fail-open (inject everywhere)
- invalid regex: stderr warning added on top of the existing try/catch fail-open
- tests trimmed to match: invalid warns, 256/257 boundary, overlong warns

51 insertions, 4 deletions across 3 files (was 299). Rebased on current main. Reopened so the thread stays intact; if you'd rather land it as a fresh PR, happy to close and resubmit.

PR #306, opened 2026-06-25
feat(adapters): add Kilo Code VS Code extension support

ousamabenyounes:
Adds Kilo Code as an instruction-tier adapter: ships `.kilocode/rules/ponytail.md` with the canonical compact body, enrolls it in `check-rule-copies.js`, and documents the project/global install paths in `docs/agent-portability.md` and the README.

The branch is rebased onto current `main`; the README conflict preserves the newer host list while adding Kilo Code. The unrelated accidental `DeepSeek-TUI` gitlink from the original commit is absent from the final diff.

## Test verification (RED → GREEN)

With the adapter files/checker enrollment absent, the focused regression test failed as expected:

```text
# tests 2
# pass 0
# fail 2
error: kilocode path missing from check-rule-copies.js
```

With the adapter applied:

```text
# tests 2
# pass 2
# fail 0
```

Full validation:

```text
node scripts/check-rule-copies.js  # pass
node scripts/check-versions.js     # pass
npm test                           # 86 root + 23 Pi + 3 MCP tests pass
diff coverage                      # 100% (28/28 changed executable lines)
GitHub Actions test                # pass on be2b9fc
```

Fix #300

PR #688, opened 2026-08-04
feat: add Kilo Code support (#300)

### LINKED ISSUE #300: [ Request ] Add support for  KILO CODE VS Code Extension
### [ Request ] Add support for  KILO CODE VS Code Extension

- Add support for the KILO CODE VS Code extension.
- Add an installation guide for the KILO CODE VS Code extension.

dhedhialy:
Fixes #300

### DIFF
--- a/.kilo/agents/ponytail.md
+++ b/.kilo/agents/ponytail.md
@@ -1,0 +1,36 @@
+---
+description: Lazy senior dev mode. Forces the simplest, shortest solution that works.
+mode: primary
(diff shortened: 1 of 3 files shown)


DietrichGebert:
Thanks! #306 covers Kilo Code with a byte-identical rule copy in .kilocode/rules/, which is the path Kilo actually reads, and it is wired into the copy checker. Going with that one, so closing this.

PR #690, opened 2026-08-04
fix: use computed settingsPath in statusline nudge (#250)

### LINKED ISSUE #250: Statusline setup nudge hardcodes ~/.claude/settings.json, ignoring CLAUDE_CONFIG_DIR

dhedhialy:
Fixes #250


DietrichGebert:
The settingsPath fix landed via #338 (f790ceb), so closing this. Your extra test assertion about CLAUDE_CONFIG_DIR could still be nice as a tiny follow-up if you want.

PR #692, opened 2026-08-04

### LINKED ISSUE #329: pi ponytail on and off issue

dhedhialy:
Fixes #329


DietrichGebert:
Thanks! We ended up solving #329 a different way: /ponytail default <mode> persists the default (bed76c5), so off can survive restarts. The mode tracker got rewritten around that design, so I'll close this one.

PR #693, opened 2026-08-04
fix: handle CRLF in loc.js fence regex and remove dead codeOf (#339)

### LINKED ISSUE #339: benchmarks: loc.js misses CRLF code fences and counts the whole response

dhedhialy:
Fixes #339

### DIFF
--- a/benchmarks/behavior.js
+++ b/benchmarks/behavior.js
@@ -10,10 +10,6 @@
 //
 // Metric: `behavior` (1 = behavior present, 0 = absent).
 
-function codeOf(text) {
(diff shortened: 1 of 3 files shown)


DietrichGebert:
Thanks! The CRLF fix landed via #518, so closing this one. If you feel like it, the one-line regression test you added here would still be welcome as a tiny follow-up PR.

PR #694, opened 2026-08-04
fix: use valid model gpt-4o-mini as default in robustness-audit.js (#399)

### LINKED ISSUE #399: benchmarks/robustness-audit.js: hardcodes invalid model gpt-5.4-mini as default

dhedhialy:
Fixes #399


DietrichGebert:
Thanks for taking a shot at #399! Turns out gpt-5.4-mini is a real, current OpenAI model, so the default works as is and the issue was closed as invalid. Closing this along with it.

PR #697, opened 2026-08-04
fix: hermes test hardcoding python3 (fails on Windows)

dhedhialy:
Fixes the Python-3 lookup in the hermes test so it passes on Windows (where 'python3' is not always on PATH).


DietrichGebert:
Thanks! This exact fix already went in via #513, so main has the python3 probe. Closing as already done.

PR #699, opened 2026-08-04
test: port additional SubagentStart matcher coverage from #522

dhedhialy:
Ports matcher coverage for SubagentStart from #522.


DietrichGebert:
Thanks for putting this together! Closing since the matcher scoping, docs and tests all landed on main via #522 (commit 1715abc), so there is nothing left for this branch to add.

PR #711, opened 2026-08-12
fix: scope the machine-global mode flag by project so concurrent sessions stop clobbering each other (#662)

### LINKED ISSUE #662: Mode flag is one machine-global file — concurrent sessions in different repos overwrite each other's mode
`ponytail-runtime.js:11-16` resolves runtime state to a single path — `getClaudeDir()/.ponytail-active` — with no repo, project or session scoping. `ponytail-subagent.js:16-21` reads only that file, and `readMode()` (`ponytail-runtime.js:28-34`) never consults `PONYTAIL_DEFAULT_MODE`. So per-repo configuration works for `SessionStart` but not for anything that reads live mode afterwards.

Two consequences when two Claude Code sessions run concurrently in different repos:

1. Repo A is configured `PONYTAIL_DEFAULT_MODE=full`, repo B `off`. A's session writes `full` to the shared file. Every subagent B spawns is then injected with the ruleset, despite B being configured `off`.
2. The reverse: starting a session in B calls `clearMode()`, deleting the flag out from under A's still-running session. A's subagents silently stop receiving the ruleset mid-session.

Both observed on a machine that routinely runs 5–10 concurrent sessions — the flag read `full` while the active repo was configured `off`.

### Repro

Sandboxed, touches no real config:

```bash
P=<plugin cache>/ponytail/4.8.4
SB=$(mktemp -d)

printf full > "$SB/.ponytail-active"                              # "repo A" turns it on
CLAUDE_CONFIG_DIR="$SB" node "$P/hooks/ponytail-subagent.js"      # "repo B" spawns a subagent
#   -> emits the full ruleset as SubagentStart additionalContext

PONYTAIL_DEFAULT_MODE=off CLAUDE_CONFIG_DIR="$SB" node "$P/hooks/ponytail-activate.js"
ls "$SB/.ponytail-active"                                          # -> gone; repo A just lost its mode
```

### Note on `PONYTAIL_SUBAGENT_MATCHER`

It can't be used to contain this. The matcher is fail-open by design (`ponytail-subagent.js:50-71`) — a missing or unparseable `agent_type`, a stdin error, or the 1s timeout all inject anyway. That's the right call for its actual purpose (scoping should never silently drop the persona), but it means it isn't a suppressor.

### Suggested direction

Not a demand, just where it seems to point: key the state file by project directory, or have `readMode()` fall back to `PONYTAIL_DEFAULT_MODE` when the reading context differs from the writing one — so per-repo config holds for subagents too.

ousamabenyounes:
## Summary

Fix #662

Live ponytail mode lives in one machine-global `.ponytail-active` file, so two Claude Code sessions in different repos clobber each other:

1. Repo A configured `full` writes the flag; a concurrent repo B configured `off` then injects the ruleset into every subagent it spawns, despite being `off`.
2. Starting a session in repo B calls `clearMode()`, deleting the flag out from under repo A's still-running session — A's subagents silently stop receiving the ruleset.

**Fix — attribute the flag to its writer.** `setMode()` records the writing project in a `.ponytail-active.owner` sidecar keyed by `CLAUDE_PROJECT_DIR` (which Claude Code sets per session). Two changes use it:

- `clearMode()` refuses to delete a flag owned by a *different* project — fixes consequence 2.
- The `SubagentStart` hook ignores the flag only when this repo defaults `off` *and* doesn't own it — fixes consequence 1.

Ownership only **whitelists**: an explicit `/ponytail` opt-in in an otherwise-off repo owns the flag and still injects, so a legitimately-active session is never suppressed. Unknown project (non-Claude hosts, which use their own per-host state dir anyway) falls back to the legacy unscoped behavior, and runtime `/ponytail` toggles keep driving subagents unchanged.

## Test verification (RED → GREEN)

New assertions in `tests/hooks.test.js` cover: an off repo ignores a foreign flag, a self-owned opt-in still injects, `clearMode` won't delete a foreign flag, and a same-repo off session still clears its own.

RED — both hooks at unmodified `main`, new tests applied:

```
# AssertionError [ERR_ASSERTION]: an off repo must ignore another repo's active flag (#662)
#     at tests/hooks.test.js:378
# pass 0
# fail 1
```

GREEN — full suite with the fix:

```
# tests 107
# pass 106
# fail 1   (pre-existing: "csv: correct pandas one-liner" needs `pip install pandas`, which CI installs)
```

`node scripts/check-rule-copies.js` and `node scripts/check-versions.js` both pass.

## Files changed

| File | Change |
|------|--------|
| `hooks/ponytail-runtime.js` | `.ponytail-active.owner` sidecar; owner-aware `clearMode`; `ownedByCurrentProject` |
| `hooks/ponytail-subagent.js` | ignore a foreign flag when this repo defaults off and doesn't own it |
| `tests/hooks.test.js` | 4 ownership scenarios for concurrent cross-repo sessions |

### DIFF
--- a/hooks/ponytail-runtime.js
+++ b/hooks/ponytail-runtime.js
@@ -4,6 +4,9 @@ const os = require('os');
 const { getClaudeDir, getConfigDir } = require('./ponytail-config');
 
 const STATE_FILE = '.ponytail-active';
+// Sidecar recording which project wrote the (machine-global) flag, so a
+// concurrent session in another repo can tell the flag isn't theirs (#662).
+const OWNER_FILE = '.ponytail-active.owner';
 
 // ponytail: VS Code Copilot never sets COPILOT_PLUGIN_DATA — it only injects
 // CLAUDE_PLUGIN_ROOT, pointed at an install path under .vscode/agent-plugins/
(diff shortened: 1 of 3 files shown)


ousamabenyounes:
Rebased onto current `main` (v4.10.0, `e3ba2aa`) to clear the merge conflict. The only collision was in `tests/hooks.test.js` — both branches added independent env-var cleanups (upstream `CURSOR_*` for #817, this PR `CLAUDE_PROJECT_DIR` for #662); kept both. No source logic changed.

Re-verified RED→GREEN locally on the rebased branch: reverting the two hook files to `main` fails `an off repo must ignore another repo's active flag (#662)` (pass 0 / fail 1); with the fix the suite passes. `check-rule-copies.js` and `check-versions.js` both pass; CI `test` green on `961fd4d`.

Now MERGEABLE. Nothing left open.

PR #722, opened 2026-08-16
feat: add Cursor plugin-tier adapter (manifest, hooks, /ponytail comm…

listepo:
…and)


Yuval-Moshe:
@listepo Any update on this one? seems really usefull for cursor users

kevinGLBdesuyo:
we need this

PR #729, opened 2026-08-18
feat(opencode): add non-breaking V2 plugin entrypoint

pascalandr:
## Summary

Add an explicit OpenCode V2 adapter targeting **OpenCode 2.0.5**, while retaining the existing V1 root entrypoint.

- Export `@dietrichgebert/ponytail/v2` and provide a local plugin directory at `.opencode/v2`.
- Register all six commands and bundled skills through the V2 plugin API; decode skills with the current `Skill.Info.path` schema.
- Persist `/ponytail lite|full|ultra|off` per session in plugin storage, including same-request switching and persistence across server restarts.
- Make bare `/ponytail` report the effective mode; invalid levels and one-shot review commands leave the selection unchanged.
- Preserve file, agent, and skill attachments while removing mention offsets invalidated by command-template expansion.
- Inject the selected rules through both `context` and `generate`, covering ordinary requests, tool continuations, and transient `session.generate` calls.
- Pin the V2 API dependency, add schema-backed tests and a focused typecheck, and install the required root dependencies in CI.
- Include an isolated-server verifier with a local mock model endpoint and Windows executable detection.

## Usage and scope

The root package export remains the V1 adapter. V2 uses the separate subpath:

```json
{ "plugins": ["@dietrichgebert/ponytail/v2"] }
```

From a checkout, run `npm ci` and configure the plugin directory:

```json
{ "plugins": ["./.opencode/v2"] }
```

Sessions without a selection use `PONYTAIL_DEFAULT_MODE` or Ponytail's `defaultMode` configuration. Subagents use their own session selection or the default rather than inheriting a parent selection. Queued mode commands update storage immediately; requests already dispatched retain their existing instructions. `off` disables plugin injection, not independently loaded `AGENTS.md` instructions. Compaction and title generation retain OpenCode's dedicated prompts.

## OpenCode 2.0.5 compatibility

The earlier beta adapter fails during skill registration on 2.0.5 because the required skill field changed from `location` to `path`. This failure was reproduced locally before integrating the fix. Transient generation now has its own hook; the verifier also uses the updated plugin activation, command, and experimental session-wait APIs.

The Windows npm package now ships `opencode.exe` while retaining `opencode2` as a command alias. The verifier detects the current executable as well as older `opencode2.exe` layouts.

## Validation

- `npm test`: all **118 tests pass**, including the pandas check and bundled subprojects.
- `npm run typecheck:v2`: passes against `@opencode/plugin@2.0.5`.
- Rule-copy, version-consistency, and diff whitespace checks pass.
- Built and installed the actual npm archive, then ran the isolated verifier on **Windows / OpenCode 2.0.5**.
- Verified command and skill registration, same-request switching, subsequent requests, session isolation, transient generation in ultra/off modes, invalid/bare commands, and persistence after restarting the private server.
- Verified Windows automatic executable detection without setting `OPENCODE_V2_BIN`.
- Runtime verification uses temporary configuration/data and a local mock provider; no paid model calls.

## Attribution

Includes contributions from @milimat0810 via https://github.com/pascalandr/ponytail/pull/1 and https://github.com/pascalandr/ponytail/pull/2, with original authorship preserved. Developed and reviewed with AI assistance.


milimat0810:
I tested this on OpenCode V2 beta-19296 and opened a follow-up against the existing PR branch: https://github.com/pascalandr/ponytail/pull/1 . It adds durable per-session mode switching, fixes stale attachment mention offsets, and corrects the local plugin directory entrypoint. The follow-up includes a repeatable isolated-server check against an installed npm archive; it verifies actual outgoing model requests and persistence across server restart. All 118 repository tests pass locally.

JosephTico:
i was using this fork and seems to have broken on 2.0.5

>The adapter targets OpenCode V2 `2.0.5`. Its skill schema requires `path`, replacing the beta API's `location` field; adapters pinned to the old beta schema fail during skill registration on 2.0.5.

milimat0810:
Fixed and verified against OpenCode 2.0.5 in a focused follow-up: https://github.com/pascalandr/ponytail/pull/2

The adapter now uses the required `Skill.Info.path`, handles the separate `generate` hook, and the verifier uses the current plugin activation, command, and session wait APIs. All 118 tests, V2 typecheck, and the isolated 2.0.5 runtime check pass.

pascalandr:
Thanks @milimat0810 for the fix! It is now in this branch: skills use `path` as required by 2.0.5. I also fixed the verifier on Windows, where the executable is now `opencode.exe` rather than `opencode2.exe`.

PR #742, opened 2026-08-21
Remove em dashes from rule copies and SKILL.md

Rashay01:
Removes em dashes (—) from prose in `AGENTS.md` and `skills/ponytail/SKILL.md`, replacing each with a comma or colon so tone/meaning is unchanged.

- `README.md` (the landing page) is untouched, as requested.
- En dashes in numeric ranges (e.g. `skills/ponytail-gain/SKILL.md`'s `6–20%`) are left alone, those are standard range notation, not the AI-tell long dash.
- Propagated the same wording fix to every exact copy of `AGENTS.md` (`.cursor/rules/ponytail.mdc`, `.windsurf`, `.clinerules`, `.agents/rules`, `.qoder/rules`, `.github/copilot-instructions.md`, `.kiro/steering`) and confirmed with `node scripts/check-rule-copies.js` that they still match byte-for-byte.
- Regenerated `.openclaw/skills/ponytail/SKILL.md` via `node scripts/build-openclaw-skills.js` to keep it in sync with the canonical `skills/ponytail/SKILL.md`.
- `npm test` passes except for the pre-existing `csv: correct pandas one-liner passes` failure, which reproduces identically on `main` before this change (missing local pandas), unrelated to this diff.

### DIFF
--- a/.agents/rules/ponytail.md
+++ b/.agents/rules/ponytail.md
@@ -14,7 +14,7 @@ Before writing any code, stop at the first rung that holds:
 
 The ladder runs after you understand the problem, not instead of it: read the task and the code it touches, trace the real flow end to end, then climb.
 
-Bug fix = root cause, not symptom: a report names a symptom. Grep every caller of the function you touch and fix the shared function once — one guard there is a smaller diff than one per caller, and patching only the path the ticket names leaves a sibling caller still broken.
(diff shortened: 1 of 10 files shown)


Rashay01:
Wrong repo, meant to target the-house-of-chai instead. Closing.

PR #744, opened 2026-08-21
chore: declare agent-plugins.org 1.0.0 manifest in root plugin.json

shashikanth171:
## Summary

- Upgrades the root `plugin.json` from a bare Claude-style stub to a valid [Agent Plugins](https://agent-plugins.org) 1.0.0 manifest (``, version, description, author, license, keywords).
- No other changes needed: `skills/` already holds all six skills in the portable Agent Skills layout, so compatible clients discover them as-is.

## Notes

- Metadata mirrors `package.json` (v4.9.0, MIT, same author/homepage).
- Zero runtime effect today — no client ships an Agent Plugins loader yet. This is free optionality for when they do; worst case it's one file to delete.
- Verified with a self-check that validates the manifest fields and discovers all six skills (`ponytail`, `ponytail-audit`, `ponytail-debt`, `ponytail-gain`, `ponytail-help`, `ponytail-review`) with matching frontmatter.

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -1,3 +1,14 @@
 {
-  "name": "ponytail"
+  "$schema": "https://agent-plugins.org/schemas/1.0.0/plugin.schema.json",
+  "name": "ponytail",
+  "version": "4.9.0",
+  "description": "Lazy senior dev mode for AI agents. The best code is the code you never wrote.",
+  "author": {
+    "name": "Dietrich Gebert",
+    "url": "https://github.com/DietrichGebert"
+  },
+  "homepage": "https://github.com/DietrichGebert/ponytail",
+  "repository": "https://github.com/DietrichGebert/ponytail",
+  "license": "MIT",
+  "keywords": ["ponytail", "lazy", "yagni", "minimal", "code-review", "skills"]
 }


shashikanth171:
## Update: Agent Plugins adoption (as of Aug 22, 2026)

This manifest is no longer speculative — the spec shipped **1.0.0 on Aug 6, 2026** (Vercel, with Amazon, Anysphere/Cursor, GitHub, Microsoft, OpenAI on the TSC) and real clients load it today.

**Compatible clients** ([agent-plugins.org/compatible-clients](https://agent-plugins.org/compatible-clients)):

| Client | Notes |
|---|---|
| OpenAI Codex / ChatGPT | v0.146.0 added manifests + workspace publishing; v0.147.0 portable install + catalog search |
| Cursor | launch partner |
| VS Code | launch partner |
| GitHub Copilot | launch partner |
| Kiro | launch partner |
| Google Agents CLI + Data Agent Kit | Google joined TSC Aug 7; more products planned |

**What this means for this PR:** all six `skills/*` are already Agent Skills-format, so the above clients can discover them from this repo's root `plugin.json` today.

**What still doesn't travel (v1 scope = skills + MCP only):** ponytail's always-on ruleset injection via hooks and the `/ponytail` slash commands stay client-specific — so the existing per-agent adapters (`.opencode/`, `.claude-plugin/`, rules copies) remain necessary for full behavior. Notably **Anthropic is absent**: Claude Code keeps its own plugin format, and **opencode isn't on the list either** — both still install via their native mechanisms.

PR #766, opened 2026-08-27
fix(copilot): defer plugin root expansion to bash

### LINKED ISSUE #759: Copilot Bash hooks resolve ${PLUGIN_ROOT} incorrectly in VS Code Remote
## Summary

The Copilot hook configuration in `copilot-hooks.json` uses `${PLUGIN_ROOT}` in its `bash` commands:

```json
"bash": "node \"${PLUGIN_ROOT}/hooks/ponytail-activate.js\"",
```

```json
"bash": "node \"${PLUGIN_ROOT}/hooks/ponytail-mode-tracker.js\"",
```

When VS Code runs on Windows and connects to a Linux remote environment (WSL), the host expands `${PLUGIN_ROOT}` before invoking Bash. The resulting command can contain a Windows-style path, which is invalid in the remote Linux shell and causes Node.js to fail to locate the hook script.

Using `$PLUGIN_ROOT` in the Bash-specific command defers expansion to Bash, which receives the correct remote environment variable.

## Environment

- Ponytail: 4.9.0 / current main
- Client: GitHub Copilot in VS Code
- Local host: Windows
- Remote extension host: Linux
- Hook configuration: `hooks/copilot-hooks.json`

## Steps to reproduce

1. Install the Ponytail Copilot plugin.
2. Open a Linux workspace using VS Code Remote from Windows.
3. Start a Copilot agent session.
4. Submit a prompt so both `sessionStart` and `userPromptSubmitted` hooks run.
5. Inspect the hook command and error output.

## Actual behavior

`${PLUGIN_ROOT}` is substituted before Bash runs. The generated command contains a host-style path or separators that are invalid in the Linux environment. Node.js cannot find `ponytail-activate.js` or `ponytail-mode-tracker.js`.

## Expected behavior

The Bash hook should resolve `PLUGIN_ROOT` using the environment of the remote Bash process and execute both scripts successfully.

## Proposed fix

Use shell-native expansion in each platform-specific field:

```diff
- "bash": "node \"${PLUGIN_ROOT}/hooks/ponytail-activate.js\"",
+ "bash": "node \"$PLUGIN_ROOT/hooks/ponytail-activate.js\"",
  "powershell": "node \"${PLUGIN_ROOT}\\hooks\\ponytail-activate.js\""
```

Apply the same change to `ponytail-mode-tracker.js`.

The PowerShell form should remain unchanged because `${PLUGIN_ROOT}` is valid PowerShell syntax. The Bash form should use `$PLUGIN_ROOT`.

This preserves the existing schema and platform behavior. An `args` field is not proposed because it is not part of the currently supported `copilot-hooks.json` command schema.

## Regression tests

Add coverage to `tests/copilot-plugin.test.js`:

1. Load `hooks/copilot-hooks.json`.
2. Assert every Bash command uses `"$PLUGIN_ROOT/..."` and does not contain `"${PLUGIN_ROOT}/..."`.
3. Assert PowerShell commands retain `"${PLUGIN_ROOT}\\..."`.
4. Execute both Bash commands with `PLUGIN_ROOT` pointing to a temporary path containing spaces.
5. Assert both processes exit with status 0.

Example runtime setup:

```js
const env = {
  ...process.env,
  PLUGIN_ROOT: pluginRootWithSpaces,
  HOME: temporaryHome,
  USERPROFILE: temporaryHome,
  PONYTAIL_DEFAULT_MODE: 'full',
};

const result = spawnSync('bash', ['-c', hook.bash], {
  env,
  input: '',
  encoding: 'utf8',
});

assert.equal(result.status, 0, result.stderr);
```

Run the complete suite with:

```bash
npm test
```

ketpatil77:
Closes #759

## Summary
- use Bash-native `$PLUGIN_ROOT` expansion for Copilot hook commands
- retain PowerShell `${PLUGIN_ROOT}` syntax unchanged
- verify platform-specific command forms and execute both Bash hooks with a plugin path containing spaces

## Verification
- `node --test tests/copilot-plugin.test.js`: 2 passed; Bash execution skipped because Bash is unavailable on this Windows host
- root `npm test`: 84 passed, 1 skipped, with one unrelated existing pandas correctness fixture failure
- `git diff --check`

### DIFF
--- a/hooks/copilot-hooks.json
+++ b/hooks/copilot-hooks.json
@@ -4,7 +4,7 @@
     "sessionStart": [
       {
         "type": "command",
-        "bash": "node \"${PLUGIN_ROOT}/hooks/ponytail-activate.js\"",
+        "bash": "node \"$PLUGIN_ROOT/hooks/ponytail-activate.js\"",
         "powershell": "node \"${PLUGIN_ROOT}\\hooks\\ponytail-activate.js\"",
         "timeoutSec": 5
       }
@@ -12,7 +12,7 @@
     "userPromptSubmitted": [
       {
         "type": "command",
-        "bash": "node \"${PLUGIN_ROOT}/hooks/ponytail-mode-tracker.js\"",
+        "bash": "node \"$PLUGIN_ROOT/hooks/ponytail-mode-tracker.js\"",
         "powershell": "node \"${PLUGIN_ROOT}\\hooks\\ponytail-mode-tracker.js\"",
         "timeoutSec": 5
       }
--- a/tests/copilot-plugin.test.js
+++ b/tests/copilot-plugin.test.js
@@ -5,7 +5,9 @@
 const test = require('node:test');
 const assert = require('node:assert/strict');
 const fs = require('fs');
+const os = require('os');
 const path = require('path');
+const { spawnSync } = require('child_process');
 
 const root = path.join(__dirname, '..');
 const REQUIRED_COMMAND_FILES = [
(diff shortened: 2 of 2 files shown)


carmona-valencia:
Thank you for addressing this so quickly in c15fb796. The change looks useful for environments where Bash receives a valid POSIX `PLUGIN_ROOT`.

I tested the commit with VS Code on Windows connected to WSL. Unfortunately, this specific case still fails because Agent Host supplies the plugin root with backslashes before Bash executes:

```text
"command": "node \"\\home\\user\\.copilot\\installed-plugins\\ponytail\\ponytail/hooks/ponytail-activate.js\""
"CLAUDE_PLUGIN_ROOT": "\\home\\user\\.copilot\\installed-plugins\\ponytail\\ponytail"
```

Node consequently resolves it below the workspace:

```text
/home/user/projects/\home\user\.copilot\installed-plugins\ponytail\ponytail/hooks/ponytail-activate.js
```

The regression test added by c15fb796 is valuable, so I do not think it should be reverted. But it supplies a valid POSIX `PLUGIN_ROOT` directly. It therefore covers spaces and Bash expansion, but not the Windows-to-WSL path conversion performed by Agent Host.

Could we treat c15fb796 as a partial fix and keep the WSL case open, possibly tracking it as an upstream VS Code Agent Host issue? The host should expose `/home/...` when hooks execute in a Linux remote extension host.

I would avoid adding WSL-specific path normalization or hard-coded installation paths to Ponytail, since that could regress native Linux, macOS, Windows, or other remote environments. The existing commit can remain useful while the cross-environment path conversion is fixed at its source.

For now, I am using a local-only override based on explicit path replacement; this resolves the warning without proposing it as portable project behavior.

```bash
ROOT="/home/user/.copilot/installed-plugins/ponytail/ponytail"
sed -i "s#\${CLAUDE_PLUGIN_ROOT}/hooks#${ROOT}/hooks#g" claude-codex-hooks.json
sed -i "s#\${PLUGIN_ROOT}/hooks#${ROOT}/hooks#g"        copilot-hooks.json
```

ketpatil77:
Thanks for testing the WSL remote case and for the concrete host output. Agreed: this PR fixes deferred Bash expansion when the host supplies a valid POSIX root, but it cannot safely normalize the backslash path emitted by VS Code Agent Host. I will keep this change narrowly scoped and treat the Windows-to-WSL conversion as an upstream host issue rather than hard-code installation paths here.

PR #771, opened 2026-08-29
refactor: shrink fallback, merge email benchmarks, dedupe bool prefs

### LINKED ISSUE #772: Hermes port: mode filter silently drops prose bullets that start with a mode word
**What**: `_filter_skill_body_for_mode` in the Hermes adapter (`__init__.py`) matches example bullets with `^-\s*([^:]+):\s*` — no quoted-value requirement. The JS filter in `hooks/ponytail-instructions.js` deliberately requires the quote (`^-\s*([^:]+):\s*"`), with a comment explaining why: worked examples are always `- lite: "..."`, and an ordinary prose bullet that happens to start with a mode word must survive verbatim.

**Effect**: under the Hermes adapter, a SKILL.md bullet like

```markdown
- Full: the ladder enforced. Stdlib and native first.
```

is misclassified as a worked example and silently dropped from the injected ruleset in every other mode. Latent today — current SKILL.md quotes all worked examples — but any future prose bullet starting with `lite`/`full`/`ultra` trips it, and only on the Hermes path, so the ports drift apart quietly.

**Repro**:

```python
body = '- Full: the ladder enforced.\n- lite: "Done, cache added."'
_filter_skill_body_for_mode(body, "ultra")
# expected: prose bullet survives, quoted lite example dropped
# actual:   both lines dropped
```

**Fix**: one-line regex change mirroring the JS guard — already included in #771 (last bullet there).

qasimsethi-code:
Ran /ponytail-audit on the repo and applied the surviving findings. Net −41 lines, no behavior changes except one deliberate fix (last bullet).

- **`getFallbackInstructions`**: condensed the 31-line hand-maintained SKILL.md mirror to paragraph form — the same shape `_fallback_instructions` in the Hermes port already uses. It only runs when `SKILL.md` (shipped in the same package) is unreadable; the full mirror was sync debt with no payoff.
- **`benchmarks/email-rate.js`** replaces `claude-email.js` + `model-email.js`: same harness loop existed twice differing only in the `call()` and env names. One file, two-entry provider table. Env contract: `PROVIDER=claude|openai`, `N`, `MODELS` (was `CE_*`/`ME_*`).
- **`.env` parsing**: hand-rolled kv parser (duplicated in both email scripts) → `util.parseEnv` (Node ≥20.12).
- **`boolPref(envVar, configKey)`** in ponytail-config folds the twin `getQuietStartup`/`getHideStatus` resolvers; exports and env→config→default semantics unchanged.
- **Hermes port fix**: `_filter_skill_body_for_mode`'s example regex was missing the `"` quote guard the JS version added deliberately — a prose bullet starting with a mode word was silently dropped in every other mode. Now mirrors `hooks/ponytail-instructions.js`.

Audit also flagged the root `plugin.json` as unreferenced — wrong, it's the Grok adapter manifest (`tests/grok-plugin.test.js` caught it); left untouched.

Tests: root 83/84 (the one failure is `csv: correct pandas one-liner passes`, environmental — no pandas on this machine, fails on a pristine checkout too), hermes 8/8, pi-extension 23/23, ponytail-mcp 3/3.

Fixes #772.


qasimsethi-code:
Friendly triage note, no rush at all: this is primarily a bugfix for #772 (the Hermes port silently drops prose bullets that start with a mode word), with the audit cleanups riding along. Since first-contribution CI needs maintainer approval here, I ran the identical test workflow on a temporary mirror — passed clean on ubuntu-latest / Node 22 / Python 3.12 (84/84 across all four suites). Happy to split the bugfix into its own minimal PR if that's easier to take.

PR #792, opened 2026-09-02
feat: add DeepSeek Harness (DSH) native plugin adapter

redtidev1918:
## Summary

Adds a native DeepSeek Harness (DSH) plugin to this repo, under `.dsh-plugin/`:

- registers the six `/ponytail*` slash commands;
- serves the six ponytail skills through the DSH skill registry;
- the plugin has zero runtime dependencies and no install scripts, and does no network, child-process, or file-write work;
- installable with `dsh plugin --profile web add` pointing at `.dsh-plugin/`; no changes to DSH core.

It is a thin adapter, not a second copy of the rules: in a checkout (or a path-based install) the provider reads the repo-root `skills/` directly — the same single source of truth every other host references. The package's own `skills/` exists only in the npm tarball (npm cannot include files outside the package directory); it is a build artifact regenerated from `../skills` by a sync script on pack/publish, and byte-checked by tests.

## Relation to #743 / #786

There are two other open DSH PRs: #743 is an older in-tree adapter built around the rc.8 integration surface (system-prompt injection and a mode-switching tool), and #786 is docs-only, pointing at the independently maintained external bundle `NOirBRight/dsh-ponytail`. This PR is a third option: a thin first-party adapter that ships with Ponytail and uses the current native command and skill registries, reusing the existing `skills/` rather than duplicating rule text. The three differ in maintenance and distribution, not in correctness — happy to close or rework this one if either of the other directions is preferred.

## Implementation

- Skills are served by a provider in the skill registry's global layer: the web profile disables the host filesystem skill discovery, so a plain skill directory would not load there. User and preset skills still override bundled ones.
- Each slash command handler drives the model with `agent.steer(...)` carrying the command briefing; a command result alone is rendered in the UI and never enters model history. This follows the built-in `/plan` command.

## Verification

- DSH plugin tests: 11/11 pass (including a drift guard that the embedded command briefings stay in sync with the shared `commands/*.toml`).
- Full repo suite: 94/95 locally (the one failure is the pandas-dependent csv test, which also fails on main without pandas; CI installs it).
- `scripts/check-versions.js` passes; `npm pack --dry-run` produces a self-contained tarball including the regenerated `skills/`.
- A headless DSH boot enumerates and loads all six skills, including with the packaged copy present and absent.

Install from a checkout:

```sh
dsh plugin --profile web add /absolute/path/to/ponytail/.dsh-plugin
```

I have not published `dsh-plugin-ponytail`; the npm name is left available if the maintainers want this distributed as a first-party package.


redtidev1918:
Reopening this so the first-party native DSH plugin approach remains available for maintainer review.

#743, #786, and this PR represent different integration and maintenance models. I'm happy to keep this implementation maintained, or rework it if the project prefers a different direction.

PR #795, opened 2026-09-02
fix: avoid shell execution when publishing OpenClaw skills

OBOneProtocol:
## Summary
- invoke `clawhub` with an argument vector instead of a shell command
- preserve normal publishing and `--dry-run` passthrough behavior
- add a regression test for repository-derived skill names containing shell syntax

## Security impact
The publishing helper previously constructed a shell command from generated skill directory names. Shell interpretation could execute content embedded in an untrusted name when a maintainer ran the release helper. Direct argument-vector execution keeps those names as literal `clawhub` arguments.

## Validation
- focused security regression
- OpenClaw publishing subsystem tests
- static syntax and patch-integrity checks
- rule-copy and version-consistency checks
- CI-equivalent root, Pi-extension, and MCP test suites
- post-validation mock trial for normal and `--dry-run` publishing

### DIFF
--- a/scripts/publish-openclaw-skills.js
+++ b/scripts/publish-openclaw-skills.js
@@ -42,26 +42,21 @@ if (slugs.length === 0) {
 const displayName = (slug) =>
   slug.split('-').map((w) => w.charAt(0).toUpperCase() + w.slice(1)).join(' ');
 
-// Minimal quoting that satisfies both POSIX sh and cmd.exe: only display names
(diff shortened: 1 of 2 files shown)


dtwkztjcmd-create (approved):

OBOneProtocol:
Hi @DietrichGebert — just checking in on this PR. It's been about a week and a half with no CI configured to auto-run, so I wanted to flag it in case it slipped through. Let me know if you'd like any changes to the fix or additional test coverage. Thanks for maintaining ponytail!

PR #799, opened 2026-09-03
fix: emit hookSpecificOutput JSON on ZCode so rules reach the model

### LINKED ISSUE #798: ZCode: SessionStart rules never injected — raw-text stdout fails ZCode's JSON validation
### Environment
- ponytail 4.9.0 (ZCode plugin marketplace)
- ZCode CLI 0.16.5, Windows 11

### Summary
On ZCode, the SessionStart hook runs — the flag file `~/.claude/.ponytail-active` is written at session start — but the ponytail ruleset never reaches the model. UserPromptSubmit context is lost the same way.

### Root cause
`writeHookOutput()` in `hooks/ponytail-runtime.js` branches per harness (Codex / Copilot / Qoder); the default branch assumes Claude Code, which injects raw SessionStart stdout. ZCode instead parses hook stdout as strict JSON and only injects `additionalContext`; plain-text output fails validation and is silently discarded.

### Repro
```sh
CLAUDE_PLUGIN_ROOT=/path/to/ponytail node hooks/ponytail-activate.js | head -c 120
# → "PONYTAIL MODE ACTIVE — level: full\n\n# Ponytail\n..."  (plain text, not JSON)
```

### Suggested fix
Add a ZCode branch emitting the JSON shape ZCode accepts — identical to the existing Qoder output:

```json
{"hookSpecificOutput": {"hookEventName": "SessionStart", "additionalContext": "..."}}
```

Note the Codex shape (top-level `systemMessage` + `hookSpecificOutput`) likely fails ZCode's strict schema too, so the Qoder shape is the safe reuse. Detection would need whatever env var ZCode exposes to plugin hooks (its Claude-compat `CLAUDE_PLUGIN_ROOT` alone can't distinguish the two).

Related: `hooks/claude-codex-hooks.json` registers a `SubagentStart` event. ZCode only supports SessionStart, UserPromptSubmit, PreToolUse, PermissionRequest, PostToolUse, PostToolUseFailure, Stop — the entry is silently dropped there. Harmless, but worth skipping on ZCode.

Arteiimis:
Fixes #798

ZCode parses hook stdout as strict JSON and injects only `additionalContext`; the native-Claude raw-text output fails validation and is silently discarded, so the ponytail ruleset never reaches the model there. The hook itself runs fine — the state flag is written at session start — which is what makes the failure so quiet.

## Changes

- Detect ZCode via `ZCODE_APP_VERSION`, which ZCode injects into every child process, hooks included. A Claude-compat variable alone (`CLAUDE_PLUGIN_ROOT`) can't distinguish the two.
- Reuse the Qoder `hookSpecificOutput` shape for ZCode in `writeHookOutput`. Unlike Qoder, ZCode has SessionStart, so `ponytail-activate.js` keeps handling startup injection — only the output shape differs from Claude Code.
- Same treatment for the `off`-mode early exit in `ponytail-activate.js`, where the bare `OK` text would also fail ZCode's strict JSON parse.
- Skip the statusline nudge on ZCode: its statusline configuration story is unverified, and the nudge points the agent at Claude's `settings.json`.
- No `SubagentStart` change: ZCode drops that event at registration time, which is already the desired behavior there.
- Tests mirror the Qoder block in `tests/hooks.test.js`: SessionStart injection, mode switch, deactivation.

## Verification

- `node --test tests/hooks.test.js` and `tests/hooks-windows.test.js` pass; `scripts/check-rule-copies.js` and `scripts/check-versions.js` pass.
- Output shape follows ZCode's documented hook contract (strict-JSON stdout, `hookSpecificOutput.hookEventName` + `additionalContext`, cf. #798 diagnostics: an event-specific output naming the wrong event fails validation).
- End-to-end injection on a live ZCode 3.10.2 session: reported in a follow-up comment once confirmed (the patch is running on the reporter's install).

### DIFF
--- a/hooks/ponytail-activate.js
+++ b/hooks/ponytail-activate.js
@@ -14,6 +14,7 @@ const {
   clearMode,
   isCodex,
   isCopilot,
+  isZcode,
   setMode,
   writeHookOutput,
 } = require('./ponytail-runtime');
@@ -26,7 +27,7 @@ const mode = getDefaultMode();
 // "off" mode — skip activation entirely, don't write flag or emit rules
 if (mode === 'off') {
   clearMode();
-  const hookOutput = (isCodex || isCopilot) ? '' : 'OK';
+  const hookOutput = (isCodex || isCopilot || isZcode) ? '' : 'OK';
   writeHookOutput('SessionStart', 'off', hookOutput);
   process.exit(0);
 }
@@ -42,7 +43,9 @@ try {
 let output = getPonytailInstructions(mode);
 
 // 3. Detect missing statusline config — nudge Claude to help set it up
-if (!isCodex && !isCopilot) try {
+// Skipped on ZCode: its statusline configuration story is unverified, and a
+// wrong pointer at Claude's settings.json would just mislead the agent.
+if (!isCodex && !isCopilot && !isZcode) try {
   let hasStatusline = false;
   if (fs.existsSync(settingsPath)) {
     // Strip UTF-8 BOM some editors prepend on Windows (breaks JSON.parse)
--- a/hooks/ponytail-runtime.js
+++ b/hooks/ponytail-runtime.js
@@ -20,6 +20,9 @@ const isCopilot = Boolean(process.env.COPILOT_PLUGIN_DATA) ||
   isVsCodeCopilotRoot(process.env.CLAUDE_PLUGIN_ROOT);
 const isCodex = !isCopilot && Boolean(process.env.PLUGIN_DATA);
 const isQoder = !isCopilot && !isCodex && Boolean(process.env.QODER_SESSION_ID);
+// ZCode injects ZCODE_APP_VERSION into every child process, hooks included.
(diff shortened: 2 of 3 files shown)


Arteiimis:
Verified end-to-end on the reporter's install (ZCode 3.10.2, Windows 11, ponytail 4.9.0 with this patch applied to the installed plugin): on a fresh session the SessionStart hook fires and the full ruleset reaches the model as injected additionalContext — no manual `/ponytail` needed. The UserPromptSubmit path (`/ponytail` mode switch) delivers its confirmation through the same envelope correctly.

ffernandodr11-art (approved):

the-loki:
Hi! Just checking in — would you happen to have any update on this PR? Thanks so much for your time!

Arteiimis:
Conflicts with `main` resolved — merged `upstream/main` (e3ba2aa), keeping both the Cursor and ZCode detection. Tests and checks pass after the merge.

PR #806, opened 2026-09-04
fix(hooks): ref the stdin fallback timer so it actually fires

### LINKED ISSUE #790: UserPromptSubmit hook still hangs to full timeout on Windows — #443 fix's setTimeout is unref'd
## Summary

The Windows stdin-EOF deadlock from #443 is not fully fixed. The mitigation added — a 1s fallback timer in `ponytail-mode-tracker.js` — never actually fires when the bug it's guarding against occurs, so the hook still hangs, just to Claude Code's external hook timeout (5000ms per `claude-codex-hooks.json`) instead of hanging forever.

## Root cause

```js
process.stdin.on('end', finish);

process.stdin.on('error', () => { finish(); process.exit(0); });
setTimeout(() => { finish(); process.exit(0); }, 1000).unref();
```

`.unref()` tells Node "don't let this timer alone keep the process alive." That's backwards for a *fallback* timer. When the `commandWindows` PowerShell wrapper swallows the piped prompt JSON (the #443 scenario) and neither `'data'` nor `'end'` ever fires, the only thing left keeping the event loop alive is `process.stdin` itself. An unref'd timer competing against a ref'd stdin handle never gets scheduled — so the 1000ms fallback is dead code in exactly the case it exists for. The process just sits until something external kills it.

## Evidence

Observed via Claude Code's own hook telemetry (`hook_cancelled` with `timedOut: true`) across a 50-session/8-day window on Windows: 2 occurrences, both recorded at exactly 5000ms — Claude Code's configured `UserPromptSubmit` hook timeout in `claude-codex-hooks.json`, not the script's intended 1000ms fallback. If the fallback timer were actually firing, these would show ~1000ms, not 5000ms.

So the bug is intermittent (stdin-swallow race, not every prompt) but when it hits, it now costs a full extra ~5s hook stall per prompt submission instead of freezing forever — better than #443, but the advertised "never hang" fallback isn't doing its job.

## Fix

Drop `.unref()` on the fallback timer (or otherwise ensure it can fire independent of stdin's ref state):

```js
setTimeout(() => { finish(); process.exit(0); }, 1000);
```

A plain (ref'd) timer will fire at 1000ms regardless of whether stdin ever emits, giving the actually-intended ~1s worst case instead of falling through to Claude Code's 5s hook watchdog.

## Environment

- OS: Windows 11
- Claude Code: 2.1.258 (native install)
- Ponytail: 4.8.4

HoneyTyagii:
## Summary

Fixes #790 (claimed issues older than 5 hours, no existing PR covers it, verified against the open PR list).

The #443 mitigation added a 1s fallback timer to the stdin hooks, but called `.unref()` on it. On Windows, when the PowerShell wrapper swallows the piped prompt JSON and neither `'data'` nor `'end'` fires, the only thing keeping the event loop alive is the ref'd stdin handle itself, and an unref'd timer competing with it never gets scheduled. The fallback was dead code in exactly the case it exists for, so the hook hung until Claude Code's external 5s watchdog killed it, observed at exactly 5000ms in hook telemetry (`hook_cancelled timedOut: true`) instead of the intended ~1000ms.

## Changes

- `hooks/ponytail-mode-tracker.js`, `hooks/ponytail-subagent.js`: drop `.unref()` on the fallback timer so it fires regardless of stdin's ref state
- `'end'` handler now exits explicitly (`finish(); process.exit(0)`) so the ref'd timer adds no latency to the normal fast path

Note: the naive fix suggested in the issue (just dropping `.unref()`) alone would have added ~1s latency to every normal hook run, because nothing exited the process on `'end'` and the then-ref'd timer kept it alive. The explicit exit on `'end'` is what keeps both properties: fast normal path, working fallback.

## Verification

- Stuck-stdin repro (spawn with pipe held open, no EOF): hook self-exits at ~1060ms with correct output instead of hanging
- Normal fast path: mode-tracker 51-54ms, subagent 53-55ms with EOF
- New regression checks in `tests/hooks.test.js`:
  - fast-path timing guard (catches the latency regression)
  - runtime hang guard via spawnSync timeout
  - static ref'd-timer contract, since the unref starvation itself is Windows-only and doesn't reproduce on Linux CI; verified this check fails on the original code and passes with the fix
- Full suite: 83/84 pass; the one failure (`csv: correct pandas one-liner`) is pre-existing on clean main, pandas isn't installed in this env, CI installs it

Fixes #790

### DIFF
--- a/hooks/ponytail-mode-tracker.js
+++ b/hooks/ponytail-mode-tracker.js
@@ -117,14 +117,21 @@ function finish() {
 }
 
 process.stdin.on('data', chunk => { input += chunk; });
-process.stdin.on('end', finish);
+// Exit on 'end', not just finish(): the fallback timer below must stay ref'd
+// (see #790) so it can actually fire when stdin is stuck, and a ref'd timer
+// would otherwise keep the process alive for its full 1000ms on this normal
+// fast path.
+process.stdin.on('end', () => { finish(); process.exit(0); });
 
 // Never hang the session. On Windows, Claude Code runs this hook through a
 // PowerShell `if {}` wrapper that can swallow the piped prompt JSON, so stdin
 // 'end' never fires and the hook blocks forever — freezing the session (#443).
 // On error, or after a short fallback, process whatever arrived (recovering the
-// mode if data came without EOF) and exit. unref() keeps the timer from adding
-// latency to the normal path, where 'end' fires first. Mirrors the best-effort,
-// never-block contract the other lifecycle hooks already follow.
+// mode if data came without EOF) and exit. The fallback timer MUST stay ref'd:
+// on Windows a stuck ref'd stdin handle keeps the event loop alive, and an
+// unref'd timer competing with it is never scheduled — so the fallback was
+// dead code in exactly the #443/#790 case it exists for, and the hook hung
+// until Claude Code's external 5s watchdog killed it (#790). Mirrors the
+// best-effort, never-block contract the other lifecycle hooks already follow.
 process.stdin.on('error', () => { finish(); process.exit(0); });
-setTimeout(() => { finish(); process.exit(0); }, 1000).unref();
+setTimeout(() => { finish(); process.exit(0); }, 1000);
--- a/hooks/ponytail-subagent.js
+++ b/hooks/ponytail-subagent.js
@@ -71,7 +71,12 @@ function finish() {
 }
 
 process.stdin.on('data', chunk => { input += chunk; });
-process.stdin.on('end', finish);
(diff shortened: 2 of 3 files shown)


codeCraft-Ritik (commented):
Just wanted to drop by and say fantastic work on this PR!

The root cause analysis on the Windows event loop .unref() starvation is brilliant, and your attention to avoiding the 1s latency regression on the fast path shows real craftsmanship. The test coverage and write-up are top-tier.

Really impressive contribution! 👏 @HoneyTyagii

Issue #817, opened 2026-09-05
Add native Cursor hooks via hooks.json


SandroPacella:
## Problem

Cursor is currently documented as an instruction-only adapter using `.cursor/rules/ponytail.mdc`. That file already sets `alwaysApply: true`, which Cursor documents as always including the rule in model context. Native hooks would add lifecycle integration and configurable mode handling; increased activation frequency over this existing installation is not established.

Users relying on automatic skill discovery have a different problem: an installed skill may never be selected. The [JetBrains evaluation](https://blog.jetbrains.com/ai/2026/07/ponytail-skill-claude-tested/) observed zero self-activations across ten Claude Code sessions with the skill merely installed. Its measured benefits came from explicitly injected Ponytail instructions. This supports reliable instruction delivery, but is not a measurement of Cursor activation or adherence.

## Proposed solution

Add native Cursor support through `.cursor/hooks.json`, reusing Ponytail's existing ruleset generation and mode logic where compatible. Validate Cursor's actual input/output contracts before adapting the existing hooks.

Cursor documents `sessionStart.additional_context` for initial context injection. Its documented `beforeSubmitPrompt` and `subagentStart` outputs do not provide equivalent context-injection fields. Event-name mapping alone therefore does not establish per-prompt or subagent injection support. Identify and verify a supported mechanism for those behaviors, and document any limitations rather than claim unverified parity.

Define how native hooks interact with the existing always-on rule so instructions are not unnecessarily duplicated and the static rule does not undermine an explicit `off` mode.

## Acceptance criteria

- [ ] Ship a valid native Cursor hooks configuration with documented installation and uninstall steps that preserve unrelated user hooks.
- [ ] Verify that a new supported Cursor session receives the configured default mode and corresponding Ponytail instructions through the documented context-injection contract.
- [ ] Validate mode changes (`lite`, `full`, `ultra`, `off`) across subsequent prompts. Implement supported behavior and explicitly document any platform limitations.
- [ ] Investigate whether the active ruleset can reach Cursor subagents through a supported mechanism. Verify delivery if supported; otherwise document the limitation rather than claim subagent parity.
- [ ] Verify coexistence or migration with `.cursor/rules/ponytail.mdc`, including avoiding duplicate instructions and honoring `off`.
- [ ] Leave one runnable compatibility check for the adapter's input/output behavior, and record a Cursor session check of instruction delivery at startup, after mode changes, and in subagents where supported.
[...]

PR #819, opened 2026-09-05
feat(cursor): add native session-start hooks

kr4chinin:
Refs #817.

Cursor currently offers fixed Ponytail guidance through an always-on rule. This adds a native `sessionStart` adapter that returns the shared, mode-filtered instructions through Cursor's documented `additional_context` field. `off` returns an empty JSON object. It leaves the Claude Code/Codex runtime unchanged.

The project installer merges `.cursor/hooks.json`, preserves unrelated hooks/settings, and supports an explicit `lite`, `full`, `ultra`, or `off` startup mode without changing other agents' defaults. Reinstall updates only this checkout's exact command; uninstall removes only that command. Installation refuses the old active `.cursor/rules/ponytail.mdc`; the migration guide preserves it as a backup outside the rules directory.

### Supported scope

- Mode changes apply to **new chats only**. Existing conversations retain their initial context; live `/ponytail` switches and per-prompt reinjection are unsupported.
- `beforeSubmitPrompt` and `subagentStart` have no documented context-injection output. `postToolUse.additional_context` misses prompts without successful tools and the start of each turn, so it is not used as a partial replacement. No automatic subagent delivery is claimed.
- Hosted cloud agents do not currently run `sessionStart`. The docs cover this restriction, other instruction copies that can undermine `off`, and the lack of an adherence comparison against the static rule.

### Validation

- `node --test tests/cursor.test.js` — passed. Exercises real adapter processes for all modes, overrides, unsupported events, malformed input, state isolation, installation/update/removal, rule migration, and preservation of unrelated/chained hooks.
- `npm test` — passed (root, pi extension, and MCP suites; Node.js 22.14.0 with pandas installed in an isolated Python environment).
- `node scripts/check-rule-copies.js` and `node scripts/check-versions.js` — passed.
- `npm pack --dry-run --json` — verified the adapter, template, installer and guide ship.
- `git diff --check` — passed.

### DIFF
--- a/hooks/ponytail-cursor.js
+++ b/hooks/ponytail-cursor.js
@@ -1,0 +1,34 @@
+#!/usr/bin/env node
+// Cursor has its own JSON contract; do not route through Claude/Codex output.
+const { getDefaultMode, normalizeMode } = require('./ponytail-config');
+const { getPonytailInstructions } = require('./ponytail-instructions');
+
+let input = '';
+let done = false;
+function finish() {
+  if (done) return;
+  done = true;
+  let output = {};
+  try {
+    const data = JSON.parse(input.replace(/^\uFEFF/, ''));
+    if (data.hook_event_name === 'sessionStart') {
+      const args = process.argv.slice(2);
+      const mode = args.length === 0 ? getDefaultMode()
+        : args.length === 2 && args[0] === '--mode' ? normalizeMode(args[1]) : null;
+      if (mode && mode !== 'off') {
+        output = { additional_context: getPonytailInstructions(mode) +
+          '\n\nCursor native adapter: the level above is fixed for this conversation. ' +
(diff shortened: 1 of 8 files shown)


hooks/ponytail-cursor.js
```diff
@@ -0,0 +1,34 @@
...
+const { getPonytailInstructions } = require('./ponytail-instructions');
+
+let input = '';
+let done = false;
+function finish() {
+  if (done) return;
+  done = true;
+  let output = {};
+  try {
+    const data = JSON.parse(input.replace(/^\uFEFF/, ''));
```
bankasimo-tech, line 13: my mom so very MoM.

bankasimo-tech (approved):

hooks/ponytail-cursor.js
kr4chinin: bro

PR #828, opened 2026-09-07
fix(pi): restore branch mode after /tree navigation

ghdtjdwn:
Navigating back to a `lite` branch with `/tree` after a later `/ponytail off` leaves Ponytail disabled and its status bar empty. Pi emits [`session_tree` after changing the selected branch](https://github.com/earendil-works/pi/blob/9767ba275f3e9a5ee0f5c5342249b629ab1b2282/packages/coding-agent/src/core/agent-session.ts#L3271-L3294), but the extension only restored persisted mode on `session_start`.

Reuse the existing branch-mode restoration for `session_tree` so the injected instructions and status bar follow the selected branch. Navigation does not append a mode entry or repeat the startup toast; a branch without a mode entry uses the configured default.

Validation:

- Added one regression test covering navigation to `lite`, `off`, and a branch without a mode entry, plus status updates and absence of history/toast side effects. It failed before the fix with `expected lite instructions` and passes afterward.
- `npm test`: 84 root tests, 24 Pi tests, and 3 MCP tests passed on macOS with Node 24.18.0, Python 3.14.7, and pandas installed in a local virtual environment.
- `node scripts/check-rule-copies.js`, `node scripts/check-versions.js`, and `git diff --check` passed.

The regression uses the existing Pi extension harness; no interactive Pi session was run.

### DIFF
--- a/pi-extension/index.js
+++ b/pi-extension/index.js
@@ -180,15 +180,23 @@ export default function ponytailExtension(pi) {
     }
   });
 
-  pi.on("session_start", async (_event, ctx) => {
+  const restoreSessionMode = (ctx) => {
     const entries = ctx?.sessionManager?.getBranch?.() || ctx?.sessionManager?.getEntries?.() || [];
-    configuredDefaultMode = getDefaultMode();
-    hideStatus = getHideStatus();
     currentMode = resolveSessionMode(entries, configuredDefaultMode);
     syncStatus(ctx);
+  };
+
+  pi.on("session_start", async (_event, ctx) => {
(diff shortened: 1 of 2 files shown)


ghdtjdwn:
The [test workflow](https://github.com/DietrichGebert/ponytail/actions/runs/34099034096) is waiting for approval. Could you approve the run and review this `/tree` mode-restoration fix when you have a chance? The regression covers returning to `lite`, `off`, and a branch using the configured default, including the corresponding status updates.

PR #831, opened 2026-09-08
fix: harden OIDC publishing dependencies and test gate

### LINKED ISSUE #825: Harden the OIDC publishing workflow against mutable dependencies
## Summary

The npm publishing workflow executes mutable dependencies while holding `id-token: write`:

https://github.com/DietrichGebert/ponytail/blob/974d940a1c5344210874150b98ff0d2c861fab6a/.github/workflows/publish.yml#L8-L24

Specifically, it uses `actions/checkout@v4`, `actions/setup-node@v4`, and installs `npm@latest`. The publish job is also not explicitly dependent on the repository test job.

Issue #435 discussed whether upgrading npm worked in the runner environment; it did not address immutable dependency pinning or OIDC privilege scope.

## Impact

A moved Action tag or newly released npm version becomes executable code in the release path without repository review. Because the same job can request an OIDC publishing token, compromise or breakage in that mutable code has direct package-release impact. A tag push can also publish without an explicit successful test-job dependency in this workflow.

## Suggested change

- Pin each Action to a full commit SHA, retaining the release tag in a comment for update tooling.
- Pin npm to an exact supported version rather than `latest`.
- Scope `id-token: write` to the publish job only.
- Make publishing depend on a passing test job, for example by exposing the existing test workflow through `workflow_call` and using `needs: test`.

Inference1:
The publishing workflow currently runs mutable Action tags and `npm@latest` with OIDC access, and a version tag can publish before tests finish. This change pins the Actions to verified release commit SHAs and npm to 11.5.1, scopes `id-token: write` to the publish job, and requires the existing test workflow to pass before publishing.

Expose `test.yml` through `workflow_call` and remove its independent tag trigger to avoid duplicate tag test runs. Pin its Actions as well and keep its permissions read-only. Add a regression check for dependency pinning, the test gate, and OIDC permission scope.

Validation:
- `npm test`: all 111 tests passed locally (Windows, Node 24.11.1).
- `node scripts/check-rule-copies.js` and `node scripts/check-versions.js`: passed.
- actionlint 1.7.12: both modified workflows passed (external shellcheck/pyflakes disabled).
- `git diff --check`: passed.
- Confirmed Action SHAs against upstream release tags and npm 11.5.1's Node engine requirements. Actual OIDC publishing was not triggered.

Closes #825.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -6,19 +6,25 @@ on:
   workflow_dispatch:
 
 permissions:
-  id-token: write
   contents: read
 
 jobs:
+  test:
+    uses: ./.github/workflows/test.yml
+
   publish:
+    needs: test
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
           node-version: '22'
       # Trusted publishing (OIDC) needs npm >= 11.5.1; Node 22 ships npm 10.
-      - run: npm install -g npm@latest
+      - run: npm install -g npm@11.5.1
       # No token: id-token: write above lets npm authenticate via OIDC, and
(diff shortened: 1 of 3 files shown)


Inference1:
@DietrichGebert Would you please review this pr

tirthposhiya1497-beep (approved):

Inference1:
@DietrichGebert Would you please review this pr

PR #859, opened 2026-09-12
fix(antigravity): bundle rules/AGENTS.md for automatic always-on context

Moussax1:
Antigravity plugin discovery looks specifically in `rules/` (e.g. `rules/AGENTS.md`) to inject always-on context.

Because `AGENTS.md` is at root and `.agents/rules/` is nested, installing as an Antigravity plugin (`agy plugin install`) registers the bundled skills but silently misses the always-on ruleset.

### Changes
- Add `rules/AGENTS.md` matching canonical rules.
- Track in `scripts/check-rule-copies.js` to guard against drift.
- Include `rules/` in `package.json` files array.
- Add smoke test in `tests/gemini-extension.test.js`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
   },
   "files": [
     "AGENTS.md",
+    "rules/",
     "hooks/",
     "skills/",
     ".opencode/",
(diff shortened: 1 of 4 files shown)


Moussax1:
Closing as adding a duplicate rules/AGENTS.md conflicts with the minimalist ethos. Better to load AGENTS.md globally in user config.

PR #864, opened 2026-09-13
opencode: support V2 plugin API alongside V1 from one entrypoint

### LINKED ISSUE #863: OpenCode 2 support: plugin uses V1-only API, does nothing on V2
## Problem

The OpenCode adapter (`.opencode/plugins/ponytail.mjs`) implements only the V1 plugin API (`config` hook, `experimental.chat.system.transform`, `command.execute.before`, `plugin` config key).

Per OpenCode's V2 docs, V1 plugin implementations do not run in V2. So on OpenCode 2 the plugin loads but never injects the ruleset or registers the `/ponytail` commands — V2 users silently get zero ponytail behavior.

## Expected

Ship a dual V1+V2 entrypoint from the same file (V1 `server()` + V2 `id`/`setup()`, per the official V1→V2 plugin migration guide) and document the V2 `plugins` config key:

```json
{ "plugins": ["@dietrichgebert/ponytail"] }
```

## Notes

- V2 mapping: system injection via a `context` session hook, commands and skills via transforms, one shared mode-flag file so switching OpenCode versions keeps the level.
- I have this working in my fork (shivamnarkar47/ponytail) with tests and can open a PR if wanted.

shivamnarkar47:
Closes #863.

## What

`.opencode/plugins/ponytail.mjs` had a V1-only default export, so on OpenCode 2 the plugin loaded but never injected the ruleset or registered commands. The default export is now one object serving both versions:

- **V1** via `server()` (object form, OpenCode >= 1.18.29) — same hooks as before; mode logic factored into shared helpers, behavior identical.
- **V2** via `id` + `setup()` — ruleset injection through a `context` session hook (agent loop only, titles/compaction/generates skip it), all 6 commands through `ctx.command.transform` with $ARGUMENTS expansion, all 6 skills through `ctx.skill.transform`, one shared mode-flag file so switching OpenCode versions keeps the level.

No new dependency (`Plugin.define` is only a type helper; the runtime duck-types `{ id, setup }`).

## Docs

README (en/es/ko) OpenCode sections show both the V1 `plugin` and V2 `plugins` keys; `docs/agent-portability.md` OpenCode row notes the dual export.

## Tests

`tests/opencode-plugin.test.js`: V1 tests ported to `server()`, plus V2 coverage with a mock context (registration, injection, `off` silence, mode persistence, template expansion) — 14/14 pass. Full suite otherwise clean (the one `csv: pandas` failure is pre-existing: pandas is not installed in this environment).

## Notes

- V1 releases older than 1.18.29 expect a function export and won't load the object form; the README's V1 snippet now implies >= 1.18.29.
- The `.mjs` file path is loaded via an explicit `plugins` entry (file entries are valid per the V2 plugins guide), not directory auto-discovery.


Zzz0zzZ0:
Opened a small PR against this branch's head with the two items noted above — V2 skills must use `Skill.Info.path` (not `location`), and a try/catch guard around the `context` hook: shivamnarkar47/ponytail#1

Merge it into `main` and #864 picks it up, or cherry-pick. No new dependency; tests included.

PR #865, opened 2026-09-13
fix(pi): expand skill alias prompts

tetienne:
## Summary

- opt in to Pi prompt-template expansion when Ponytail slash aliases send `/skill:...` messages
- preserve follow-up delivery while the agent is busy
- cover idle and busy alias dispatch

Pi 0.84.0 defaults `sendUserMessage(...).expandPromptTemplates` to false, so without this option the model receives the literal skill command instead of the skill body.

## Verification

- `TMPDIR=/tmp node --test pi-extension/test/extension.test.js`

The root suite additionally requires the documented local pandas prerequisite, which is unavailable in the sandbox.

### DIFF
--- a/pi-extension/index.js
+++ b/pi-extension/index.js
@@ -103,12 +103,12 @@ export default function ponytailExtension(pi) {
     const message = normalized ? `${skillName} ${normalized}` : skillName;
 
     if (ctx?.isIdle?.() === false) {
-      pi.sendUserMessage(message, { deliverAs: "followUp" });
+      pi.sendUserMessage(message, { deliverAs: "followUp", expandPromptTemplates: true });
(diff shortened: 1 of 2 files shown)


fitchmultz:
Verified at `40aa9cfe5ac4b201c1411e6b316041f60d962eef`:

- `node --test pi-extension/test/extension.test.js`: 12/12 pass.
- Loaded this PR's actual extension through Pi's native loader on published Pi 0.85.1 and upstream `e4ce7b449`.
- Invoked `/ponytail-review` while idle and during an active turn. Both paths delivered the expanded skill body to a scripted provider; the busy path retained follow-up delivery. No real model calls were made.

This fixes the literal `/skill:ponytail-review` reaching the model that I reproduced without these options. The native expansion flag is sufficient; no extension-side skill parsing is needed. Thanks for the fix.

PR #867, opened 2026-09-13
fix: silence pi "description is required" warnings from root doc files

tkoyama010:
## Problem

When ponytail is installed as a [pi](https://github.com/badlogic/pi-mono) package, pi treats direct `.md` children of the package root as skill candidates. `AGENTS.md`, `README.md`, `README.es.md`, `README.ko.md`, and `after-install.md` each produce a `description is required` warning under the **[Skill conflicts]** banner on every pi launch.

## Fix

List the root doc files in `.fdignore` — read by pi and fd, but **not** by ripgrep, so the docs stay searchable.

Two gotchas handled:
- `.ignore` would also hide them from ripgrep, so `.fdignore` is used instead.
- Glob patterns like `/*.md` don't work: pi strips the leading `/` from ignore patterns, degrading `/*.md` to `*.md`, which matches every `SKILL.md` in the repo. Hence explicit filenames.

## Verification

```
pi skill loader on this repo: 6 skills, 0 diagnostics (was: 5 warnings)
```


tkoyama010:
Closing: incorrect premise. pi only scans a package's skills/ directory for SKILL.md files plus direct .md children of skills/ itself — it never treats package-root docs (README.md, AGENTS.md) as skill candidates, so no warnings arise from them. Verified against pi's DefaultPackageManager.resolve().

Issue #683, opened 2026-08-03
ponytail-audit: no pass to flag AI-shaped test brittleness (fixture-name branches in prod, snapshot/mock-only assertions, 1:1 method-mirror tests)


wdi-dave-roberts:
ponytail's test stance is minimal and construction-time: "non-trivial logic leaves one runnable check." But `ponytail-audit`, scanning existing code, has no pass to flag AI-generated *test* slop, which is a distinct and common failure mode:

- tests that mirror implementation methods 1:1 (break on any refactor; they assert structure, not behavior)
- snapshot-only or mock-call-only assertions (they pin the AI's own output, not observable behavior)
- worst: fixture-name / id / timestamp branches leaking into *production* code to make a test pass

These pass CI and expose no deletable abstraction, so the ladder skips them, but they're maintainability slop, and the fixture-branch-in-prod case is also a correctness smell (production behavior special-cased for test data).

Distinct from #602, which is about ponytail constraining the agent *while writing* tests; this is the *audit* skill *detecting* brittleness in tests that already exist.

Suggested: one `## Hunt` bullet: fixture-name/id/timestamp branches in production code (encode the real invariant instead); snapshot-only / mock-call-only assertions (prefer observable behavior); tests that mirror methods 1:1 rather than behaviors.

Source: Google Testing Blog, "Test Behaviors, Not Methods."

Happy to open the PR.

PR #881, opened 2026-09-16
feat: add reuse tag to ponytail-review and ponytail-audit (#868)

### LINKED ISSUE #868: Feature: add a `reuse` tag to ponytail-review/audit for in-repo duplication (ladder rung 2 at review time)
## Background

The ladder used at write time (`skills/ponytail/SKILL.md`) has rung 2, "Already in this codebase? Reuse it, don't re-write it." (added via #253, following #217).

But `ponytail-review` and `ponytail-audit` — the skills that check code *after* it's written, on a diff or on the whole repo — only have five tags: `delete`, `stdlib`, `native`, `yagni`, `shrink`. None of them cover "this new helper duplicates one that already exists elsewhere in this same repo." So the write-time rung has no corresponding check-time tag: if a helper gets duplicated anyway, review/audit currently has nothing to catch it with.

## Observed case

We had a session that self-reported adding 4 small helpers, although equivalent helpers already existed in the same repo (a Python/JS monorepo). Running the current review tags over that diff would not have flagged it — none of `delete/stdlib/native/yagni/shrink` matches "equivalent code already exists elsewhere in this repo."

## Proposal

Add a sixth tag to both `skills/ponytail-review/SKILL.md` and `skills/ponytail-audit/SKILL.md`:

- `reuse:` an equivalent function/script/config already exists in this repo at `<path>`. Replacement: point to it.

And add one line to the instructions in both skills: "Search the same repo for an existing equivalent before flagging or accepting a new helper."

## Local workaround

Until then, we append one line to the review prompt ourselves: "Search the same repo for an existing equivalent (function/script/config); if found, list it as reuse."

Happy to send a PR if the direction is welcome.

nolanchic:
Fixes #868.

The write-time ladder tells the agent to reuse what's already in the repo (rung 2), but the check-time skills had no matching tag — a diff that adds a fourth helper for something `src/lib/` already ships sailed through `delete/stdlib/native/yagni/shrink` untouched. The audit side had the same blind spot, repo-wide.

This adds the sixth tag to both skills, slotted after `native:` so the three "it already exists somewhere" tags read as a group — stdlib ships it, the platform ships it, this repo ships it. The replacement is just a path. Review also gets the instruction line the issue proposed (search for an existing equivalent before flagging or accepting a new helper, framed as rung 2 checked after the fact) and a worked example in the same one-line format; audit gets the corresponding Hunt entry.

OpenClaw copies regenerated via `build-openclaw-skills.js`. Guard scripts pass; `npm test` is 94/95 locally — the one failure is the pandas CSV correctness check, which fails identically on a clean checkout here (no local pandas) and is green in CI where pandas is installed.

One deliberate omission: the compact rule copies (`AGENTS.md` and the seven host copies) are untouched — the review tags live only in the two skills, so `check-rule-copies.js` stays quiet and the diff stays minimal.


kitchen7c:
Heads-up: I'm preparing a PR for #683 (ponytail-audit hunts AI-shaped test brittleness) that adds a paragraph to the same `## Hunt` section of `skills/ponytail-audit/SKILL.md` (+ regenerated `.openclaw` copy) that this PR touches. Textually adjacent only — no logical dependency. Proposed landing order: this PR first, #683 rebases if needed; happy either way, just flag it here.

PR #883, opened 2026-09-18
chore(skill): use the compact fallback text as the skill body

tieubao:
The SessionStart hook injects this file every session; the author's fallback text carries the same rules in 2.7k chars instead of 6.6k.


tieubao:
Opened against the wrong base by mistake (meant for my fork). Closing, sorry for the noise.

PR #889, opened 2026-09-19
fix: let projects that ban code comments skip the ponytail: marker (#887)

### LINKED ISSUE #887: The mandated `ponytail:` comment has no opt-out, and conflicts with codebases that ban code comments
## Summary

The skill mandates a `ponytail:` code comment with no way to opt out. In a codebase whose style guide bans code comments, that rule is a direct conflict, and the agent resolves it in favour of ponytail because the skill is loaded on every turn while the project's style guide is one line among many.

I am not asking for the marker to be removed. `ponytail-debt` greps for it, so it is load-bearing rather than decorative, and the debt ledger is one of the better ideas in this plugin. I am asking for a supported way to keep the laziness and put the ceiling somewhere other than the source file.

## The rule

`skills/ponytail/SKILL.md:64`:

> Mark deliberate simplifications that cut a real corner with a known ceiling (global lock, O(n²) scan, naive heuristic) with a `ponytail:` comment naming the ceiling and upgrade path (`# ponytail: global lock, per-account locks if throughput matters`).

The same sentence ships in eleven other files, so a user who edits one still gets the rule from the rest:

- `AGENTS.md:28`
- `.agents/rules/ponytail.md:28`
- `.clinerules/ponytail.md:28`
- `.github/copilot-instructions.md:28`
- `.kiro/steering/ponytail.md:33`
- `.qoder/rules/ponytail.md:28`
- `.windsurf/rules/ponytail.md:28`
- `.openclaw/skills/ponytail/SKILL.md:52`
- `skills/ponytail-debt/SKILL.md:4,11,35`
- `.openclaw/skills/ponytail-debt/SKILL.md:8,32`
- `.opencode/command/ponytail-debt.md:5`

## Why it is a problem rather than a preference

Plenty of codebases treat a code comment as a defect. The position is not fringe. The `dhh` skill in the 37signals skill set, calibrated against roughly 200 of DHH's real review comments on basecamp/fizzy, lists under **Question These**:

> Comments that say what, not why — "It says what's happening, but not why?"

A `ponytail:` marker is usually a why, so it survives that particular bar. It does not survive a stricter house rule that says the reason for a change belongs in the commit message or the pull request, not beside the code. Under that rule the marker is exactly the thing being banned, and the plugin offers no switch.

The practical effect is that an agent running with ponytail writes markers, a later cleanup pass removes them as policy violations, and the next ponytail turn writes them back. The user is the only thing breaking the loop, repeatedly.

## What would fix it

Any one of these, in order of preference:

1. **An off switch.** An env var or a level flag, for example `PONYTAIL_MARKERS=off`, that turns the mandate into "name the ceiling and its upgrade path in the commit message or the PR body instead". The ladder, the rules and the output format are unaffected.
2. **Defer to the project.** Add a clause to the rule: when the project's own style guide or agent instructions ban code comments, record the ceiling in the commit message and skip the marker. Cheap, one sentence, and it matches the plugin's own stated precedence that user instructions win.
3. **A ledger file.** Write markers to a tracked file, say `.ponytail-debt.md`, keyed by `file:line`. `ponytail-debt` reads that instead of grepping. This keeps the ledger working in codebases that will not accept comments at all, at the cost of drift when lines move.

Option 2 is the smallest change and would be enough for me. Option 1 is the most explicit. Option 3 is the most work and probably not worth it unless others ask.

Happy to open a PR for 1 or 2 if you tell me which shape you would take.

ShantanuTaro:
Hi! Thanks @viktorianer for writing up #887 so clearly. The loop you described (ponytail adds a marker, the cleanup pass deletes it, the next turn adds it back) is annoying, and I liked that you already had the smallest fix in mind.

I went with your option 2. It's the one sentence you suggested, added to the end of the marker rule:

> …with a `ponytail:` comment naming the ceiling and upgrade path. **If the project's own instructions ban code comments, put that note in the commit message or PR body instead.**

This keeps the part that matters: the agent still writes down the ceiling and the upgrade path. It just puts that note in a place the codebase allows. It also fits how the plugin already works, since project and user instructions take priority. Codebases that allow comments see no change, so `ponytail-debt` keeps working for them.

**What changed**
- `skills/ponytail/SKILL.md` and `AGENTS.md`
- the 7 other copies of the rules that `scripts/check-rule-copies.js` keeps in sync (Cursor, Windsurf, Cline, Qoder, `.agents`, Copilot, Kiro)
- `.openclaw/skills/ponytail/SKILL.md`, regenerated with `node scripts/build-openclaw-skills.js`

**How I checked it**
- `node scripts/check-rule-copies.js` passes: all copies match `AGENTS.md` and all 9 invariants are still there.
- In `npm test`, everything passes except `csv: correct pandas one-liner passes`. That test fails the same way on `main` on my machine, and it looks like it's just missing a local pandas install.

I left options 1 (an env var) and 3 (a ledger file) out for now, since the issue said option 2 would be enough. I'm happy to add either one if you'd like the explicit switch as well.

Closes #887


viktorianer (approved):

viktorianer:
Great. Thanks. 👍🏼

PR #907, opened 2026-09-19
feat: add OpenCode V2 plugin support with V1 fallback

### LINKED ISSUE #892: plugin: OpenCode V2 incompatible - "Plugin must export a default definition with an id and an effect or setup function"
## Summary

The `@dietrichgebert/ponytail` plugin (v4.10.0) fails to load on OpenCode V2 because it exports a function instead of the V2-required `Plugin.define()` object shape.

## Environment

- opencode version: 2.0.8
- OS: Linux 7.2.6-1-cachyos (linux x64)
- Active plugins: @dietrichgebert/ponytail@4.10.0, superpowers

## Reproduction

1. Install `@dietrichgebert/ponytail` globally via OpenCode V2
2. Add `"plugin": ["@dietrichgebert/ponytail"]` to `~/.config/opencode/opencode.json`
3. Start OpenCode — plugin fails to load on every startup

## Expected Behavior

The plugin loads successfully and injects the ponytail ruleset into system prompts.

## Actual Behavior

The plugin fails to load with:

Plugin must export a default definition with an id and an effect or setup function.
(cause: SchemaError(Expected object at "default"))

The current entrypoint (`ponytail.mjs`) exports a function:

```js
export default async ({ client } = {}) => {
  return { config, 'experimental.chat.system.transform', 'command.execute.before' }
}
OpenCode V2 expects a Plugin.define() call:
export default Plugin.define({
  id: "ponytail",
  async setup(ctx) { ... }
})
Additional Context
- Plugin was designed for OpenCode V1's plugin API
- The plugins config field (plural) is now the V2 equivalent of the old plugin (singular)
- Full V2 plugin docs: https://opencode.ai/v2/docs/build/plugins

### PR DESCRIPTION
**Pull request by RubenPari**
## Summary

Migrate the Ponytail plugin to support OpenCode V2 while maintaining OpenCode V1 compatibility.

Fixes #892: "Plugin must export a default definition with an id and an effect or setup function"

## Changes

- **V2 Plugin API**: Export `Plugin.define({ id: 'ponytail', async setup(context) })` as default
- **V1 Fallback**: Keep `pluginV1` named export for legacy loaders
- **Shared Implementation**: Extract hook logic into `createHooks()` to eliminate duplication
- **Config Update**: Add `"plugins"` field to `opencode.json` for V2 compatibility
- **Test Coverage**: Expand test suite to cover both V1 and V2 APIs (16 tests total, all passing)

## Hook Behavior

All three hooks remain unchanged:
- `config` — registers slash commands and skills directory  
- `experimental.chat.system.transform` — injects ponytail ruleset into system prompts
- `command.execute.before` — persists `/ponytail` mode switches

## Testing

- ✅ 6 V1 API tests (function-based plugin)
- ✅ 6 V2 API tests (Plugin.define object shape)
- ✅ 4 utility tests (command parsing, plugin shape validation)
- ✅ Full test suite passes with no regressions

## OpenCode Setup

**V1**: Add to `opencode.json`: `{ "plugin": ["./.opencode/plugins/ponytail.mjs"] }`
**V2**: Add to `opencode.json`: `{ "plugins": ["./.opencode/plugins/ponytail.mjs"] }`

Both are now supported in the config.

## Files Changed

- `.opencode/plugins/ponytail.mjs` — Refactored for dual API support
- `opencode.json` — Added V2 plugins field
- `tests/opencode-plugin.test.js` — Expanded test suite for both APIs

### CODE REVIEW

**Comment by sjnims** [comment]
Thanks for taking on the V2 migration — the direction is right: switching the default export to an `{ id, setup }` object does satisfy the V2 loader schema and makes the `Plugin must export a default definition with an id and an effect or setup function` startup error go away.

However, there is one blocking issue: **as written, this loads cleanly on OpenCode 2.x but silently does nothing.** That's arguably worse than the current loud failure.

## `setup()` must not return the hooks object

Verified against the published `@opencode/plugin` 2.0.10 type definitions (`dist/promise/plugin.d.ts`):

```ts
export interface Plugin {
  readonly id: string;
  readonly setup: (context: Context) => Promise<Cleanup | void> | Cleanup | void;
  //                                          Cleanup = () => Promise<void> | void
}
```

The return value of `setup` is a **cleanup function** (or nothing) — not a hooks map. Hooks are registered via the context object:

- system-prompt injection → `ctx.session.hook("context", (event) => event.system.push({ type: "text", text }))`
- slash commands → `ctx.command.transform((editor) => editor.add({ name: "ponytail", execute }))`
- skills → `ctx.skill.transform(...)`
- events → `ctx.event.subscribe(...)`
- tool hooks → `ctx.tool.hook("execute.before" | "execute.after", ...)`

The V1 hook keys used in `createHooks()` (`config`, `experimental.chat.system.transform`, `command.execute.before`) do not exist in the V2 host surface, so returning `createHooks(client)` from `setup` is ignored: no ruleset injection, no `/ponytail` command, no mode persistence — and no warning to tell you.

## Smaller issues

1. **`context.client` doesn't exist in V2.** The context exposes `app`, `location`, `session`, `command`, `tool`, `event`, `plugin`, etc. The `context?.client || context` fallback binds to the context itself, so any client-dependent behavior silently degrades.
2. **The tests are circular.** They call `pluginV2.setup({})` by hand and then invoke the returned V1 hook keys directly. They never exercise the V2 runtime contract, so the 16 passing tests don't demonstrate V2 compatibility. A mock of the real `Context` shape (with `session.hook`, `command.transform`, …) would have caught this.
3. **`"plugins"` is not a config key.** The published `opencode.ai/config.json` schema defines only `plugin` (singular), and `Config` has `additionalProperties: false`, so the plural field is ignored at best.

## Suggested shape

This is the verified working pattern for a V2 server plugin (same one a known-good plugin in the wild uses):

```js
import { Plugin } from "@opencode/plugin";

export default Plugin.define({
  id: "ponytail",
  async setup(ctx) {
    ctx.session.hook("context", (event) => {
      const mode = readMode();
      if (mode === "off") return;
      event.system.push({ type: "text", text: getPonytailInstructions(mode) });
    });

    ctx.command.transform((editor) => {
      editor.add({
        name: "ponytail",
        description: "Switch ponytail mode (lite|full|ultra|off)",
        execute: async (inv) => { /* parse args, writeMode(...) */ },
      });
    });

    // skills via ctx.skill.transform(...)
  },
});
```

If dual V1/V2 support is a goal, separate entrypoints via the package `exports` map (root = V2 definition, e.g. `./plugin-v1` = legacy function) would be safer than one dual-shape file — V1 loaders treated every exported function as a plugin (as this PR's own test comments note), and how they handle a non-function `default` is untested here.

Happy to be corrected if a newer runtime reintroduces hooks-object returns from `setup`, but against 2.0.10's published types this needs a rework before merge.

